PR #2, opened 2020-01-07
Use ChaCha RNG

vinc:
Use ChaCha RNG seeded by RDRAND.


vinc:
Can't compile this PR at the moment because of that: https://github.com/rust-lang/rust/issues/61721

PR #8, opened 2020-01-20
Improve editor

vinc:
- [x] Newline should split the current line in half
- [x] Backspace at beginning of a line should merge it with the previous one
- [x] Implement scrolling


vinc:
A bug while scrolling down has been temporary fixed by disabling dual vga and serial output.

PR #16, opened 2020-02-04
Add PhysBuf for DMA

vinc:
When we allocate on the heap, the next three frames after the first allocated frame are used to create new page tables, and are therefor unavailable.

If we allocate a buffer larger that the space available in the first frame, it will then use the fifth frame, but if we use the buffer for DMA the frames used for the page tables will be overwritten.

https://docs.rs/x86_64/0.8.3/x86_64/structures/paging/mapper/trait.Mapper.html#tymethod.map_to
https://docs.rs/x86_64/0.8.3/src/x86_64/structures/paging/mapper/mapped_page_table.rs.html#92

### DIFF
--- a/src/kernel/allocator.rs
+++ b/src/kernel/allocator.rs
@@ -3,10 +3,9 @@ use alloc::vec::Vec;
 use core::ops::{Index, IndexMut};
 use crate::kernel;
 use linked_list_allocator::LockedHeap;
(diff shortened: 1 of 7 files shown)


vinc:
For now a naive solution is to reallocate if the space is not contiguous in physical memory.

PR #18, opened 2020-02-07
Better shell history

vinc:
This PR will make it possible to persist the shell command history between sessions and improve the transition to and from history mode.

The `up` and `down` keys are used to navigate the history while any other key will replace the current command by the command currently shown from the history. The `down` key can go one step beyond the history vector to show the current command.

### DIFF
--- a/src/user/shell.rs
+++ b/src/user/shell.rs
@@ -163,6 +163,10 @@ impl Shell {
         }
     }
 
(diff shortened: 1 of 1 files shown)


vinc:
Right now the path to the file history is hardcoded to `/usr/admin/.shell_history` rather than `~/.shell_history` because the OS still don't have a process struct to store the current user.

PR #19, opened 2020-02-07
Add process struct

vinc:
Add a process struct to store the current dir, user, and env vars.

This will allow the use of relative paths and of a home dir set in env vars.

The `go` command, equivalent to the `cd` in Unix and Windows, can now be used. With `go <dir>` to change the current dir and `go` without argument to print the current dir.

### DIFF
--- a/src/kernel/fs.rs
+++ b/src/kernel/fs.rs
@@ -34,9 +34,9 @@ pub fn realpath(pathname: &str) -> String {
     if pathname.starts_with("/") {
         pathname.into()
     } else {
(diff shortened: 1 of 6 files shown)


vinc:
At the moment there is only one process but this will change.

vinc:
Use `goto` instead of `go` for the fullname command, and still use `g` for its short name. It could of course be aliased to `cd`.

```
> goto /tmp

> list
alice.txt
hello.txt

> read hello.txt
Hello, world!

```

PR #111, opened 2020-11-07
Add ACPI shutdown

vinc:
Implement ACPI shutdown instead of using QEMU shutdown.

See https://wiki.osdev.org/Shutdown#ACPI


vinc:
Tested successfully on QEMU, Bochs, and my laptop. AML parsing fails on VirtualBox but I can't figure out why so for now I hardcoded the value needed. Will have to go back and fix the issue.

PR #112, opened 2020-11-08
Add new date command

vinc:
Back when MOROS was using `cargo-xbuild` we couldn't compile the `time` crate (https://github.com/time-rs/time/issues/268) but now that it doesn't use it anymore (https://github.com/vinc/moros/pull/83) we can compile it.

So we can now have a new `date` command that accepts formatting (https://time-rs.github.io/time/time/index.html#formatting).

```
> date
2020-11-08T23:47:02

> date "%A, %B %d, %Y"
Sunday, November 08, 2020
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Add new `date` and `env` commands (#112)
 - Add ACPI shutdown (#111)
 - Improve text editor (#109)
 - Add pcnet driver (#82)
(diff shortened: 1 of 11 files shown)


vinc:
With nanoseconds precision:

```
> date "%F %H:%M:%S.%N"
2020-11-09 00:20:17.307953119
```

vinc:
Added an `env` command to set a `TZ` env var used by the `date` command to display the date with a UTC offset.

In the future we'll want to import a time zone database to find the UTC offset for a given date and time zone, we could also add a geosearch to the database to use latitude and longitude instead of strings like `Europe/Paris` and be compatible with the `geotime` command that use longitude.

At the moment MOROS will continue to be generally unaware of time zones and assume the RTC is using locale time.

PR #113, opened 2020-11-10
Add new `mem` command

vinc:


vinc:
The command only show the heap utilization at the moment:
```
> mem
Size: 1048576
Used:    4008
Free: 1044568
```

vinc:
We also increased the heap size to prevent a crash when the PCNET driver and the ACPI tables could both fit in memory.

PR #114, opened 2020-11-11
Add a web server

vinc:


vinc:
The last couple of commits are kind of like reinventing FTP over HTTP but at the moment our FUSE driver is read only and this is quite simple to write to live update the file system:

```
HTTP Server listening on 0.0.0.0:80
10.0.2.2 - - [11/Nov/2020:18:36:43 +0000] "GET /tmp/hello.txt" 404 20
10.0.2.2 - - [11/Nov/2020:18:36:58 +0000] "PUT /tmp/hello.txt" 200 0
10.0.2.2 - - [11/Nov/2020:18:37:02 +0000] "GET /tmp/hello.txt" 200 15
10.0.2.2 - - [11/Nov/2020:18:37:09 +0000] "DELETE /tmp/hello.txt" 200 0
10.0.2.2 - - [11/Nov/2020:18:37:19 +0000] "GET /tmp/hello.txt" 404 20
```

vinc:
The server has a few major limitations that will need to be addressed later:
- Concurrency is not supported
- Requests larger than the MTU are not supported
- Chroot is not supported

Concurrency will be implemented when threads are available, and the second one might be handled at the network driver level.

The file transfer protocol is therefor limited to files below the MTU size, and while we can read, write, and delete files, we can only read and write directories, it's not currently possible to delete them. This part of the code should probably be moved to a proper FTP or TFTP server.

Finally we should support the concept of `index.html` files to manually replace automatic directory listing.

PR #118, opened 2020-11-14
Add tests

vinc:
It's about time we added some tests right?

See: https://github.com/phil-opp/blog_os/issues/591

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: rust
+cache: cargo
+rust:
+  - nightly
+install:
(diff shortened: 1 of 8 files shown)


vinc:
Compiling `rand_chacha` in test fail with the following error:
```
LLVM ERROR: Do not know how to split the result of this operator!
```
This seems related to https://github.com/rust-lang/rust/issues/61721 that prevented us to build the crate in release until a few days ago, see https://github.com/vinc/moros/pull/116

PR #196, opened 2021-06-28
Add syscalls

vinc:
Add the following syscalls:
- [x] Sleep
- [x] Uptime
- [x] Realtime


vinc:
This PR add one function that takes an argument and two functions that returns an argument. Later we'll add more syscalls but first we need to change quite a few things in the kernel.

PR #213, opened 2021-07-25
Fix floating point comparisons

vinc:
See: https://rust-lang.github.io/rust-clippy/master/#float_cmp


vinc:
This first fix suggested by clippy is actually bad: https://floating-point-gui.de/errors/comparison/

vinc:
So instead we'll use the `float-cmp` crate: https://crates.io/crates/float-cmp

vinc:
Using the default for `approx_eq!` might not work for all numbers, but it should be good enough for now.

PR #216, opened 2021-07-26
Update crypto crates

vinc:
We couldn't update the crypto crates for a long while because of the following LLVM error when compiling `sha2`:

```
LLVM ERROR: Do not know how to split the result of this operator!
```

But a solution for that was merged here https://github.com/RustCrypto/hashes/pull/203

Also see https://github.com/RustCrypto/hashes/issues/202


vinc:
The latest version of `rand_chacha` compile without error on my computer but on the CI it gets the same LLVM error as `sha2`.

PR #221, opened 2021-07-27
Refactor editor

vinc:
- [x] Replace calls to `vga` functions by CSI codes
- [x] Simplify line editing by copying new code from `Prompt` trait


vinc:
The editor now work with the serial output.

PR #222, opened 2021-07-29
Add a regular expression engine

vinc:
Add a very basic regular expression engine based on https://www.cs.princeton.edu/courses/archive/spr09/cos333/beautiful.html

- [x] Support `.` char
- [x] Support `*` char
- [x] Support `+` char
- [x] Support `?` char
- [x] Support `\` char
- [ ] Support `|` char
- [x] Support `\d` chars
- [x] Support `\D` chars
- [x] Support `\s` chars
- [x] Support `\S` chars
- [x] Support `\w` chars
- [x] Support `\W` chars
- [ ] Support `{a,b}` range
- [x] Support lazy `?` char

Also add a `find --line` command that works like `grep` on Unix to test it on files.

### DIFF
--- a/doc/regex.md
+++ b/doc/regex.md
@@ -1,0 +1,21 @@
+# MOROS Regular Expression Engine
+
+MOROS include a simplified regular expression engine with the following syntax:
+
+- `\` escape the following character to its literal meaning
+- `^` matches the starting position within the string
+- `$` matches the ending position within the string
+- `*` matches the preceding element zero or more times
+- `+` matches the preceding element one or more times
+- `?` matches the preceding element zero or one time
+- `.` matches any single character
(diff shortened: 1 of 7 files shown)


vinc:
The first command using the regex engine will be `find` used to find files with a given name or lines in files with a given pattern:

```sh
> find /tmp --name ".*txt" --line "exa\w+ple"
```

vinc:
Finding `Alice` and `cat` in the example text:
find

Now we probably need to add greedy matching to find `cats?`

vinc:
The `find --name` part of the `find` command will be added in another PR with the `Wildcard` struct.

vinc:
Recursive line matching:
find

PR #228, opened 2021-08-05
Add basic userspace

vinc:
This PR is an experiment into implementing a userspace mode.

I started from this article: https://nfil.dev/kernel/rust/coding/rust-kernel-to-userspace-and-back/
And the OSDev wiki as usual: https://wiki.osdev.org/Getting_to_Ring_3
Then I found this gem: https://github.com/WartaPoirier-corp/ananos/blob/dev/docs/notes/context-switch.md
Along with the associated code in the repo that is a perfect starting point to get an idea of how to get something working.


vinc:
To produce the binary of an example program to run in userspace, first create a file `prog.asm` with something like this:
```asm
bits 64
start:
  mov rax, 0x0
  mov rdi, 0x3ff0000000000000
  mov rsi, 0x0
  mov rdx, 0x0
  int 0x80
  jmp start
```
Then run the following command:
```sh
nasm -f bin prog.asm -o prog.bin && xxd -i prog.bin
```
And you will get this byte array:
```rust
  0xb8, 0x00, 0x00, 0x00, 0x00, 0x48, 0xbf, 0x00, 0x00, 0x00, 0x00, 0x00,
  0x00, 0xf0, 0x3f, 0xbe, 0x00, 0x00, 0x00, 0x00, 0xba, 0x00, 0x00, 0x00,
  0x00, 0xcd, 0x80, 0xeb, 0xe3
```
In MOROS this program will call `sleep(1.0)` through a syscall repetitively.

vinc:
We can use `include_bytes!("../dsk/bin/sleep")` to skip the manual conversion into a byte array. But since we have a filesytem the next step is to read from the disk at runtime.

vinc:
Userspace \o/
userspace

vinc:
The debug outputs from the above screenshot are now replaced by a program `/bin/test` that send a number to the syscall `TEST` who print it, increment it, and send it back. The program will also increment what it receive, sleep 1 second, then start again.

This should suffice to show communication between userspace and the kernel.

vinc:
The string in the hello world binary is at address 32 in the code but accessing it result in a page fault:

page-fault

vinc:
The address from a user code segment just needed to be translated by the kernel to fix the exception above.

vinc:
Next step will be to add an ELF loader.

PR #248, opened 2021-09-24
Add ELF loader

### LINKED ISSUE #249: Freeze when typing
There seems to be a bug introduced recently that freeze the OS when typing sometimes. It may be related to change to the interrupts during sleep.

vinc:
This PR will bring us closer to be being able to run rust binaries, but some issues are still remaining. For now we'll be able to run ELF binaries produced in the following way:

```
nasm dsk/src/bin/hello.s -o /tmp/hello.o -f elf64
ld -o dsk/bin/hello /tmp/hello.o -Ttext=100 -Tdata=200
```

Instead of raw binaries produced like so:

```
nasm dsk/src/bin/hello.s -o /tmp/hello -f bin
```


vinc:
The previous commits should fix #249

PR #252, opened 2021-09-29
Add dynamical disk information

vinc:
- [x] Add `SuperBlock` struct
- [x] Add `LinkedBlock` struct
- [x] Add `Drive` struct
- [x] Store total and allocated block count in super block
- [x] Dynamically allocate the fs based on disk size or memory size
- [x] Dynamically allocate the heap based on memory size
- [x] Use the whole block for bitmap blocks
- [x] Remove hardcoded bitmap and data area and size
- [ ] Add dynamical offset to superblock area

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Add dynamical disk information (#252)
 - Add ELF loader (#248)
 - Add basic userspace (#228)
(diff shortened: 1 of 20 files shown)


vinc:
Next step: rewrite the `BlockBitmap` to use `Block` instead of `LinkedBlock` and make full use of the data buffer instead of skipping the first 4 bytes containing the next block address.

PR #253, opened 2021-10-02
Remove array-macro dependency

sugar700:
Hello, I'm a maintainer of `array-macro` crate.

Code that was using `array_macro::array!` used it incorrectly causing a subtle issue. In 2.0.0 the behaviour of `array` macro did change to clone values to match the behavior of standard library's [`vec! macro`](https://doc.rust-lang.org/stable/std/macro.vec.html).

In the code that used this macro it was used for objects of `PhysBuf` type.

```rust
#[derive(Clone)]
pub struct PhysBuf {
    buf: Arc<Mutex<Vec<u8>>>,
}
```

For `Arc` objects cloning doesn't create a new object, rather it shares a reference causing a subtle issue where all elements of `rx_buffers` shared the exact same buffer, which I don't think you have intended.

Standard library now contains `array::map` which can replace `array_macro::array!` in most situations. It may be a good idea to use that, considering amount of people that look at standard library's code is greater than number of people that look at `array_macro` (which is just me).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,12 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "array-macro"
-version = "2.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a08c8f180019011a1d94e5eebcbfaa548736815f636fadc10955f1b622b4c71"
-
-[[package]]
 name = "arrayvec"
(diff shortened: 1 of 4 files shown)


vinc (approved):

vinc:
Wow I didn't notice the issue but with your explanation it's very clear, thanks a lot 👍

PR #254, opened 2021-10-02
Enable array_map feature

vinc:
This was required for #253


sugar700:
You may want to update your nightly as array::map was stabilized three months ago, but sure. In fact, the GitHub Actions logs make it clear:

```
warning: the feature `array_map` has been stable since 1.55.0 and no longer requires an attribute to enable
 --> src/lib.rs:5:12
  |
5 | #![feature(array_map)]
  |            ^^^^^^^^^
  |
  = note: `#[warn(stable_features)]` on by default
```

PR #258, opened 2021-10-22
Fix VGA issues with real hardware

vinc:
### During VGA initialization
- [x] Clear screen
- [x] Disabled underline
- [x] Set default VGA palette registers
- [x] Set default VGA colors

### After boot
- [x] Enable cursor


vinc:
It seems that coreboot+seabios doesn't set the VGA palette registers to the same defaults as the regular BIOS so we need to set during initialization and load the default colors. We could display the register during boot to be sure of that.

https://stackoverflow.com/a/64427761/6999496
https://github.com/coreboot/coreboot/blob/4.14/src/drivers/pc80/vga/vga.c#L66
https://github.com/coreboot/seabios/blob/rel-1.14.0/vgasrc/stdvga.c#L58

PR #261, opened 2021-10-29
Add website

vinc:
Add a static website with a simple landing page hosted at [moros.cc](http://moros.cc).

The server allow both HTTP and HTTPS to be browsable from inside MOROS so the website can be used in request examples.


vinc:
Next step will be to host the website from MOROS with the httpd server running in a cloud virtual machine with QEMU.

PR #262, opened 2021-10-31
Add shell redirections

vinc:
The initial plan for MOROS Shell was to emulate redirections with pipes and `read` or `write` commands like `print hello -> write /tmp/test` that can be shortened to `p hello > w /tmp/test` where `>` is the equivalent of `|` for Unix shells.

Pipes are not yet implemented but with a `DUP` syscall and the file handle table it's easy to implement redirections like `print hello => /tmp/test` where `=>` is the same than `>` for Unix shells.


vinc:
For now only `print hello => /tmp/test` or `print hello 1=> /tmp/test` works while `read <= /tmp/test` for example wouldn't  because reading from stdin is not implemented for most tools.

PR #268, opened 2021-11-09
Add process table and exit syscall

vinc:
Improve process structure and use a table to save them.

This will allow use to save the registers of a process before spawning a new one so we can switch back to it later.


vinc:
Debug message in the EXIT syscall:
process

vinc:
Two issues:
- We can't boot with 32 file handles per process, so we have to carefully manage 16 file handles
- The boot is too slow with the `release` flag so the tests fail

vinc:
Finally got `exit` working so we can run a program with `spawn` and go back to the shell, but somehow it only work with the debug output activated.

vinc:
We can now run a program multiple times and unmap the memory after use, well except sometime the unmapping doesn't work. And sometime the compiled code trigger a reset.

vinc:
The new exit syscall work well to return to the kernel shell for flat binaries, but some issues remains for elf binaries that will be addressed next.

userspace

PR #276, opened 2021-11-28
Add pci command and switch IDE controllers to compatible mode

vinc:


vinc:
Output of the new `pci` command:

pci

PR #303, opened 2022-01-12
Add directory support to userspace

vinc:
- [x] Create dirs from userspace
- [x] Delete dirs from userspace
- [x] Read dirs from userspace
- [x] Add `delete` syscall
- [x] Add `/dev/null`
- [x] Add `proc` command
- [x] Rename `stat` syscall to `info`
- [x] Rename `FileStat` struct to `FileInfo`


vinc:
The following pages were interesting when implementing the `read` syscall for directories:

- https://www.bo-yang.net/2011/03/07/plan9-intro-note-4
- http://man.cat-v.org/plan_9/5/read

In MOROS the `stat` syscall on a file or a directory returns the file attributes (called `FileStat`) that is a subset of its directory entry for userspace programs. The `read` syscall on a directory returns the same structure for each entry until its buffer cannot fit the next entry.

vinc:
Rename `stat` syscall and `FileStat` to `info` and `FileInfo`.

Issue #323, opened 2022-04-10
Executing a user process breaks network statistics


vinc:
```
> net stat
rx: 2 packets (1180 bytes)
tx: 2 packets (620 bytes)

> /bin/hello
Hello, World!

> net stat
rx: 36254196902658239 packets (16192026705920 bytes)
tx: 6278066737626506568 packets (7390416128 bytes)
```

PR #327, opened 2022-04-11
Fix issues with process alloc

vinc:


vinc:
Adding the offset to the frame allocator should fix #323

PR #329, opened 2022-04-15
Fix  device reading

vinc:


vinc:
We could read devices in `find` but it will wait for an input when reading the console so we can't `find /` without hanging unexpectedly. Reading a device file is not useful at the moment in `find` so we disable it.

vinc:
We could read devices in `find` but it will wait for an input when reading the console so we can't do `find /` without hanging unexpectedly. Reading a device file is not useful at the moment in `find` so we disable it.

PR #348, opened 2022-06-07
Add shell variables

vinc:
- Add access to the the pathname of the script or the shell via `$0`
- Add access to script arguments via `$1`, `$2`, `$3`, and so on
- Add access to environment variables like `$HOME`
- Allow setting a shell variable with `test = 42` to access it via `$test`

### DIFF
--- a/doc/shell.md
+++ b/doc/shell.md
@@ -121,3 +121,25 @@ Which is more efficient than doing:
 
     > print hello -> write /tmp/hello
 
+## Variables
+
+- Name of the shell or the script: `$0`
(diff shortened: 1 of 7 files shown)


vinc:
The shell will create a var for each env var at launch but a subsequent call to `env test=42` for example will not make a `$test` available like `test = 42` would because `env` is not a built-in command at the moment.

vinc:
Should we transform `test = 42` into `set test = 42` (or `let test = 42`) and add `unset test`?

PR #349, opened 2022-06-08
Executable loading

vinc:
Trying to spawn a regular file would generate an exception so this PR add a header to binary file to detect them during the syscall and return an error in case the detection failed that will be used by the shell.

Just like ELF files that start with `7F 45 4C 46` (the last bytes coding for `ELF` in ASCII), the flat binary used by MOROS at the moment will now start with `7F 42 49 4E` (the last bytes for `BIN`).


vinc:
This last commit actually fixed ELF loading! But the PR is introducing a new bug because `/bin/hello` can't be run more than once without cause a general protection fault.

PR #350, opened 2022-06-10
Add binaries

vinc:
- [x] Add userspace binaries for commands that don't allocate nor read command line args
- [x] Add binaries in `/bin` to command autocomplete


vinc:
The userspace programs that only use syscalls and don't try to allocate or read command line arguments can be moved from `src/usr` to `src/bin` and run in ring 3.

PR #351, opened 2022-06-12
Read command line args from userspace programs

vinc:
- Add args to SPAWN syscall
- Store args in the memory of the programs and pass them to their entry points
- Rewrite `print` and `sleep` as userspace programs


vinc:
WIP:

```
> hello world test
DEBUG: args_len:   2
DEBUG: args_ptr:   0x444444442330 (src)
DEBUG: args_ptr:   0x444444440468 (tmp)
DEBUG: args_ptr:   0x4444454c0009 (dst)
DEBUG: code addr:  0x444445440000
DEBUG: heap addr:  0x4444454c0000
DEBUG: stack addr: 0x444445540000
DEBUG: args:       0x80009 (len=2)
DEBUG: args[0]:    0x80000 (len=5)
DEBUG: args[1]:    0x80005 (len=4)
Hello, World!
arg: DEBUG: SYSCALL WRITE (handle=1, ptr=0x0 (0x444445440000), len=0)

arg: DEBUG: SYSCALL WRITE (handle=1, ptr=0x0 (0x444445440000), len=0)
```

We can pass a slice of str to a userspace program and iterate it but cannot read the individual str. From the kernel side we translated their pointers to userspace addresses, but from the user side the pointers are null.

PR #352, opened 2022-06-15
Add shell globbing

vinc:
Add support to the shell of filename expansion or globbing for `*` and `?` wildcard characters, where a pattern given in an argument of a command will be replaced by files matching the pattern.

- `*` means zero or more chars except `/`
- `?` means any char except `/`

For example `/tmp/*.txt` will match any files with the `txt` extension inside `/tmp`, and `a?c.txt` will match a file named `abc.txt`.


vinc:
Only the `delete` command support multiple filenames as arguments for now.

PR #357, opened 2022-06-20
Add shell aliases

vinc:


vinc:
Example of shell session:

```shell
> /tmp

> env
TZ    "7200"
DIR   "/tmp"
HOME  "/usr/vinc"

> env FOO "Hello, World!" # Set env var and shell var

> env FOO
Hello, World!

> env
TZ    "7200"
DIR   "/tmp"
HOME  "/usr/vinc"
FOO   "Hello, World!"

> print $FOO
Hello, World!

> set foo "Hi there" # Set shell var

> alias p "print"

> p $foo
Hi there

> unset foo # Unset shell var

> unset FOO # Unset env var and shell var

> unalias p # Remove command alias
```

TODO: `env` will have to be reimplemented as a builtin command to set shell variables, and `$DIR` is not yet an env var.

vinc:
Some aliases are present in `/ini/shell.sh` but commented out to offer an experience closer to Unix. This would add `ls /tmp` instead of `list /tmp` or it's alias `l /tmp` for example.

Some commands could also be removed, for example `list /tmp` is equivalent to `read /tmp`, the only differences being the error message when trying to read a file or device, and automatically listing the current directory when invoked without arguments. Another case is `goto /tmp`, aliased to `g /tmp` and `go /tmp` that has a shorter equivalent with simply `/tmp`. And finally `quit` with its alias `q` have become the main way to quit a program so `exit` could be moved to the Unix compatibility aliases because the shortcut `e` is already taken by `edit`.

PR #360, opened 2022-06-26
Use exit code to set status var

vinc:
Should a program exit or return with a `isize`, `usize`, `ExitCode`, `Result<(), usize>`, or `Result<(), ExitCode>`. A few options were tried in this PR to end up on the latter, but we may simplify it again as it was before to `ExitCode`.

At least now we can get the code with `print $?` in the shell, and the userspace binaries no longer need to return anything by default instead of `0`.


vinc:
Some tests with `print foo => /bar` are failing when using the RAM FS in this PR and the previous one with no obvious fix: https://github.com/vinc/moros/pull/357

Calling the console to print an empty string with the serial driver consistently fix them, but calling the serial driver directly doesn't work.

vinc:
With this last commit we go back to the usual `0` for success and `> 0` for failure.

- https://en.wikipedia.org/wiki/Exit_status
- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02
- https://www.freebsd.org/cgi/man.cgi?query=sysexits&sektion=3

PR #365, opened 2022-07-05
Improve http command

vinc:
- [x] Fix error when the host arg is not the first arg
- [x] Add support for binary output


vinc:
This PR is another step to install the system from the network instead of including every files inside the binary, we'll be able to do the following:

    > http moros.cc/download/files/0.7.1/bin/sleep => /bin/sleep

PR #388, opened 2022-08-14
ATA: Sata LBA2 wrong address

ToguyC:
According to [this osdev page](https://wiki.osdev.org/ATA_PIO_Mode#Detecting_device_types), the address for the CH register (LBA2) for the SATA drives should be 0xC3 instead of 0x3C.


vinc (approved):

vinc:
Hi @ToguyC, good catch, thanks for the fix!

PR #399, opened 2022-08-27
Update shell redirections

vinc:
> "Those who don't understand Unix are condemned to reinvent it, poorly."

I've been reading the doc I wrote about file handle redirections in the shell and they were associated with the thin arrow `->` (pipes) instead of the fat arrow `=>` (redirections).

While I was fixing that I noticed that I didn't have a use for a  `<` shortcut because `>` was used for thin arrows instead of fat arrows. Fat arrows can be written in both directions, so it'd make more sens to have both shortcuts for them, like in Unix. And then `|` became really tempting as a shortcut for the thin arrow, like in Unix. I resisted that in the past, instead preferring to use it for the or-combinator. If I use the keyword `or` for the or-combinator (or even `||` as in Unix) then `|` is free for the pipe shortcut...

We'll see, but now `>` will stand for `=>` and `<` for `<=`.

While I was doing that I also revisited the decision in #387 to have `=>` appending to files, instead I made it truncate first as it was before and added `=>>` which can be shortened to `>>` ... like in Unix. It's really useful to have both append and truncate in one command instead of having to use `delete` first for the latter. I was missing it while I tested the feature.

In this PR I'm also adding a long syntax for file handles in redirections with the explicit `[1]=>` being equivalent to the implicit `=>` while keeping the short `1=>` that can now be even shorter with `1>`, making it even shorter than in Unix. `[1]=>[3]` will also be possible in the future.

### DIFF
--- a/doc/shell.md
+++ b/doc/shell.md
@@ -132,10 +132,8 @@ NOTE: The following file handles are available when a process is created:
 - `stderr(2)`
 - `stdnull(3)`
 
-<!--
-NOTE: A redirection with a fat arrow will append to a file without truncating
-it first. This could change in the future with with the addition of `=>>`.
--->
+A redirection with a single arrow head will truncate its destination while
+multiple heads like `=>>` will append to it.
 
 NOTE: Arrows can be longer, and also shorter in the case of fat arrows:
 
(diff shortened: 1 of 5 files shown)


vinc:
All the following commands are equivalents:

    > time read foo.txt --> write /dev/null
    > time read foo.txt -> write /dev/null
    > t r foo.txt | w /dev/null # ?

    > time read foo.txt ==> /dev/null
    > time read foo.txt => /dev/null
    > t r foo.txt > /dev/null

    > time read foo.txt [1]=> /dev/null
    > t r foo.txt 1> /dev/null

    > time read foo.txt [1]=>[3]
    > t r foo.txt 1>3

PR #400, opened 2022-08-28
Add RTC device file

vinc:
- [x] Add a device file for the RTC at `/dev/rtc`
- [x] Allow writing to the device to change the date and time

### DIFF
--- a/src/api/clock.rs
+++ b/src/api/clock.rs
@@ -5,6 +5,10 @@ pub const DATE_TIME_ZONE: &str = "%Y-%m-%d %H:%M:%S %z";
 pub const DATE_TIME: &str = "%Y-%m-%d %H:%M:%S";
 pub const DATE: &str = "%Y-%m-%d";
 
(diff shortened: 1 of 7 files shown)


vinc:
```
> read /dev/rtc
2022-08-28 09:23:21

> lisp
MOROS Lisp v0.3.0

> (load "/lib/lisp/core.lsp")
true

> (read-file "/dev/rtc")
"2022-08-28 09:24:47"
```

vinc:
Some RTC have a century register at `0x32` but it could be at another port defined in an ACPI table. It can also be absent, or worse, wrong. I experimented with it but had issues with dates in future centuries.

For now dates should be between the beginning of 2000 and the end of 2099:
```
> print "2099-12-31 23:59:57" => /dev/rtc

> date
2099-12-31 23:59:58 +0000

> date
2099-12-31 23:59:59 +0000

> date
2000-01-01 00:00:00 +0000
```

Hopefully we'll have a better system by the end of the century :)

PR #401, opened 2022-08-28
Add more bytes to RNG seed when RDRAND is not available

vinc:


vinc:
The feature can be tested with `-cpu core2duo` in QEMU

PR #449, opened 2022-12-01
Improve lisp

vinc:
Add improvements to the Lisp to be able to participate in the Advent of Code 2022.

- [x] Add `chunks` function
- [x] Add `slice` function
- [x] Add `sort` function
- [x] Add `uniq` function
- [x] Add `filter` function
- [x] Add `intersection` function
- [x] Add `trim` function
- [x] Add `split` function
- [x] Add `chars` function
- [x] Add `words` function
- [x] Add `splice` with `@` shortcut in variadic functions
- [x] Change `lines` function
- [x] Change `length` function
- [x] Change `set` function


vinc:
The `sort_by` used in `sort` had to be changed to `sort_unstable_by` has it was causing a general protection fault while running the tests of the shell.

vinc:
Splitting the code of `default_env` reduced the kernel file size from 2008064 to 2000896 bytes. This is extremely close to the 2 MB limit where the kernel will overwrite beginning of the rest of the file system.

vinc:
The function `(set <symbol> <value>)` will now return the value instead of the symbol, unlike `(define <symbol> <value>)` that still returns the symbol.

```
(define i 0)    # => i
(set i (+ i 1)) # => 1
(set i (+ i 1)) # => 2
```

vinc:
Add `splice` with `@` as a shortcut for variadic functions:

```
((function (a @b) a) 1 2 3) # => 1
((function (a @b) b) 1 2 3) # => (2 3)
```

PR #454, opened 2022-12-12
Add process page table

vinc:
For now the process page table is just a copy of the kernel page table, but this PR adds code to switch to the new page table during the `spawn` syscall and switch back to the previous page table during the `exit` syscall.


vinc:
There was an alloc error on the first call of a program in userspace at address `0x44444563b190` which is `process.code_addr + process.entry_point_addr`:

```
~
> print hello
DEBUG: Could not map Page[4KiB](0x44444563b000) to PhysFrame[4KiB](0x1480000)
Error: Could not allocate address 0x44444563b190

~
> print hello
hello
```

Which was fixed by allocating a page at `code_address` without waiting for the first page fault that would allocate it.

vinc:
This PR was closed just one day short of its first anniversary :tada:

PR #478, opened 2023-04-14
Refactor lisp functions

vinc:
- [x] Rename `print` to `p` and `println` to `print`
- [x] Rename `string-join` to `join-string` (aliased to `join-str`)
- [x] Use `atom?` instead of `atom` by default
- [x] Use `equal?` instead of `eq` by default
- [x] Use `variable` (aliased to `var`) instead of `define`
- [x] Keep `define` shortcut (aliased to `def`) for `define-function`
- [x] Add `str`, `str->num`, `str->bin`, `bin->str` aliases
- [x] Add `num-type`, `num->bin`, `bin->num` aliases
- [x] Add `uniq`, `rev`, and `len` aliases
- [x] Add `bool?`, `str?`, `sym?`, `num?`, `fun?`, `mac?` aliases
- [x] Use short aliases in core, doc, and examples


vinc:
This change of keyword syntax will make the language a bit more in line with something like Ruby or JavaScript which might be easier for beginners using the OS. And the aliases make it shorter. The syntax is pretty close to the Janet language that I discovered this week!

```janet
(load "/lib/lisp/core.lsp")

(def (foobar n) (do
  (var i 1)
  (while (< i n) (do
    (if (and (= (% i 3) 0) (= (% i 5) 0))
      (print "foobar")
      (if (= (% i 3) 0)
        (print "foo")
        (if (= (% i 5) 0)
          (print "bar")
          (print i))))
    (set i (+ i 1))))))

(foobar 100)
```

It'd be great to add an implicit `do` to `def` and `while`.

PR #495, opened 2023-06-13
Update rust version

vinc:


vinc:
Before the update we couldn't do two `print hello` in a row, and after the update it work again. We had a similar problem that was temporarily fixed the same way the last time.

vinc:
But the test that uses `print` is now failing like it was before.

PR #501, opened 2023-06-13
Fix lisp cond expand

vinc:
With the following macro for `eq?`
```janet
(var eq?
  (macro args `(equal? ,@args)))
```
The macro was not expanded while expanding `cond`
```
> (cond ((equal? "a" "a") "oh noes"))
"oh noes"

> (cond ((eq? "a" "a") "oh noes"))
Error: Expected first argument to be a function
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


vinc:
There might be a more general solution during eval to expand any macros left.

PR #506, opened 2023-06-17
Improve chess game

vinc:
- [x] Add `save` command
- [x] Add `load` command
- [x] Add `play` command
- [x] Improve `perf` command output
- [x] Improve error output

chess-load

### DIFF
--- a/src/usr/calc.rs
+++ b/src/usr/calc.rs
@@ -142,6 +142,12 @@ fn repl() -> Result<(), ExitCode> {
 }
 
 pub fn main(args: &[&str]) -> Result<(), ExitCode> {
+    for &arg in args {
+        match arg {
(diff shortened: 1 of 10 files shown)


vinc:
The save format uses the standard [Forsyth–Edwards Notation](https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation).

It would have been better to use the [Portable Game Notation](https://en.wikipedia.org/wiki/Portable_Game_Notation) to save the history of the game but the chess engine requires `std` to load this format. I may modify the engine to use `nom` internally instead of `regex` to work with `no_std` at some point.

chess-hex

The `perf` chess command uses the same kind of output as the `hex` shell command.

chess-perf

PR #511, opened 2023-06-29
Add namespaces to lisp

vinc:
This PR introduces a lower level `file` namespace to the lisp allowing reading (`file:read`) and writing (`file:write`) directly from a file handle to have a better `print` function so the output of a script can now be redirected to a file from the shell for example.

This wasn't possible before:
```
lisp /tmp/lisp/pi.lsp 100 => pi.txt
```

The new notation using a dot `.` to group functions within a namespace will also be used for `number`, `string`, and `regex` with `number.type`, `string.trim`, and `regex.match?` for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Add namespaces to lisp ([#511](https://github.com/vinc/moros/pull/511))
 - Update smoltcp from 0.9.1 to 0.10.0 ([#510](https://github.com/vinc/moros/pull/510))
 
 ## 0.10.0 (2023-06-21)
--- a/doc/lisp.md
+++ b/doc/lisp.md
@@ -196,3 +196,6 @@ language and reading from the filesystem.
 - Add full support for line and inline comments
 - Add params to function representations
 - Add docstring to functions
+
+### Unreleased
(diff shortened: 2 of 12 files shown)


vinc:
New escaped tab `\t` and carriage return `\r` with the existing newline `\n`:

```
> (print "Hello,\tWorld!")
Hello,	World!
()

> (print "Hello,\rWorld!")
World!
()

> (print "Hello,\nWorld!")
Hello,
World!
()
```

vinc:
```
> (mod -1 3)
2

> (rem -1 3)
-1

> (% -1 3)
-1
```

vinc:
There are a few options for the namespace separator:
1. `file::open`
2. `file:open`
3. `file.open`
4. `file/open`

The first option `::` is very common but so far I was thinking about using the second one `:` that is shorter. Both could be slightly confusing if I introduce keywords starting with a colon like `:foo` though.

A slash `/` would borrow the directory metaphor but might be confusing if the language is used as a shell whereas a dot `.` would work like it does for hostnames and filenames, for example: `http.get "example.com/foo.txt"`.

vinc:
If the language is used as a shell it might be interesting to rename `read-file`, `write-file`, and `append-file` to `read`, `write`, and `append` but the latter already exists.

PR #512, opened 2023-07-07
Add network syscalls

vinc:
- [x] Add `/dev/net/tcp` device file to open a TCP socket handle
- [x] Add `/dev/net/udp` device file to open a UDP socket handle
- [x] Add `poll` syscall
- [x] Add `connect` syscall
- [x] Add `listen` syscall
- [x] Add `accept` syscall
- [x] Rewrite `tcp` command to use the network using syscalls only
- [x] Rewrite `http` command to use the network using syscalls only
- [ ] Rewrite `httpd` command to use the network using syscalls only
- [x] Rewrite `socket` command to use the network using syscalls only
- [x] Rewrite `host` command to use the network using syscalls only
- [ ] Rewrite `dhcp` command to use the network using syscalls only


vinc:
Syscalls used for a TCP Client:

```
- OPEN /dev/net/tcp -> handle
- CONNECT handle address port
- WRITE handle buffer -> bytes written
- READ handle buffer -> bytes read
- CLOSE handle
```

vinc:
We could introduce an equivalent of Unix `IOCTL` named `CONTROL` or `MANAGE` to replace `CONNECT handle address port` with `MANAGE handle request response` where `request` could be `"connect 192.168.1.1 80"` for example.

vinc:
We need a `poll` syscall to be able to wait on multiple handles without blocking.

For example:

```rust
let list = vec![(stdin, Event::Read), (socket, Event::Read)];
loop {
  if let Some((h, e)) = syscall::poll(&list) {
    // Read from stdin or socket, whichever is ready first
  }
}
```

vinc:
The updated `socket` command using `poll` allows us to have a chat interface with other computers.

The command `socket -l 1234` and `socket 192.168.1.100 1234` have the equivalent on Unix with `nc` and can be used to launch a basic chat server/client on port 1234.

What's missing is knowing when a socket connection is closed. We could add another `IO::Close` event to poll.

vinc:
Poll is not blocking for now, so we could poll only one file handle at a time instead of a list. The list is only useful to reduce the number of userspace/kernelspace boundary crossing. But if poll was blocking the we would cross it only once at the beginning. And in that case we need the list because we can't poll the file handles one by one.

vinc:
With a blocking `poll` the console and socket `read` don't need to be blocking.

vinc:
Sometime the TCP state after a `connect` change directly from `syn-sent` to `close-wait` without seemingly going through `established` but that's only because the transition is too fast. In that case we can check if there's something that can be read to know if an error occurred or not.

vinc:
There was some confusion with the syntax for the port number between `tcp` and `socket` on one side and `http` on the other side:

- `tcp <host> <port>`
- `socket [<host>] <port>`
- `http <host>[:<port>][ ]<path>`

We can't change `http` because of the path so this commit changes the first two commands to remove the confusion.

- `tcp <host>:<port>`
- `socket [<host>:]port`
- `http <host>[:<port>][ ]<path>`

We don't support IPv6 yet but we might want to use `!` instead of `:` like Plan 9 does when this will be the case.

vinc:
Finally we also introduced a special case in `read` to return the socket state without blocking when the given buffer is 1 byte. This is a simple hack that we are also using for reading the console where 4 bytes is `read_char` and something else like 256 is `read_line`.

vinc:
There's two options for DHCPv4, either using the `smoltcp` crate's implementation of the protocol in the kernel by creating a special `dhcp` socket, or reimplementing the protocol in userspace using a pair of UDP sockets. The latter is the most interesting way to do it but this work could be done later in another PR.

The HTTP server will also have to be rewritten but this can be done in a separate PR too.

vinc:
The socket file size will now show the recommended buffer size:
```
> list /dev/net
1446 2023-07-29 12:09:21 tcp
1458 2023-07-29 12:09:21 udp
```
Size: MTU - Ethernet header - IP header - TCP/UDP header

vinc:
We can now delete a device file. This was needed to easily recreate the socket files and change their sizes without recreating the file system.

PR #516, opened 2023-07-19
Speed up VGA text mode

vinc:
Scrolling was a bit slow in QEMU and a simple refactor of the new line method made it appear almost instantaneous.

Before:
```
> time read /tmp/alice.txt
...
Executed 'read /tmp/alice.txt' in 1.691047s
```

After:
```
> time read /tmp/alice.txt
...
Executed 'read /tmp/alice.txt' in 0.003999s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


vinc:
This second refactor doesn't bring any measurable difference but the resulting code is smaller and clearer.

PR #521, opened 2023-07-24
Fix cursor movements on long lines in editor

vinc:
This PR fix a couple of bugs happening on lines larger than 80 or 160 chars in the editor during cursor movements.

### DIFF
--- a/src/usr/editor.rs
+++ b/src/usr/editor.rs
@@ -236,25 +236,24 @@ impl Editor {
         self.highlighted.clear();
     }
 
-    // Align cursor that is past the end of line,
-    // to the end of line or to the left of the screen.
+    // Align cursor that is past the end of the line, to the end
+    // of the line or to the left of the screen.
     //
-    // If the cursor is at the end of the long line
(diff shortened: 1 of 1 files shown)


vinc:
Moving right past the end of the screen (that is 80 chars wide) on a line that was longer than 160 chars would make the cursor jump 80 chars too far on the right and disappear after 160 chars (2x80). Same issue when going left after 160 chars. The issue wasn't happening when adding chars to the line.

vinc:
Moving from a long line of more than 80 chars to a shorter line but still more than 80 chars would not move the cursor to the end of the line, or to the left of the screen if the line is not visible, but instead the cursor would move 80 chars past the end of the line.

vinc:
Found another bug: when we want to move the cursor from the end of a long line (> 80 chars) to a line that is shorter than the current screen offset, the editor will move the cursor to the left of the screen, but if we insert a char there, it's past the end of the line and will cause a panic. We need to change the editor to move the cursor to the end of the shorter line and update the offset.

PR #542, opened 2023-11-12
Add more code examples to lisp

vinc:
- [x] NTP client
- [x] Fetch client

### DIFF
--- a/dsk/tmp/lisp/fetch.lsp
+++ b/dsk/tmp/lisp/fetch.lsp
@@ -1,0 +1,15 @@
+# HTTP/0.9 Proxy (https://github.com/vinc/fetch)
+
+(load "/lib/lisp/core.lsp")
+
+(var proxy-host "10.0.2.2")
+(var proxy-port 8888)
+
+(var stdout 1)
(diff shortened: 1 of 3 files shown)


vinc:
The NTP code is inspired by https://seriot.ch/projects/tiny_ntp_client.html

```
~
> host pool.ntp.org
5.196.8.113

~
> lisp /tmp/lisp/ntp.lsp 5.196.8.113
1699815734
```

vinc:
The last commit adds a client to [Fetch](https://github.com/vinc/fetch), a HTTP/0.9 server and proxy to other protocols like HTTPS and Gemini.

```
~
> lisp /tmp/lisp/fetch.lsp https://example.com
<!doctype html>
<html>
<head>
    <title>Example Domain</title>

    <meta charset="utf-8" />
    <meta http-equiv="Content-type" content="text/html; charset=utf-8" />
```

With the following running on the host system if MOROS is running on QEMU:

```
~
> ruby fetchd.rb --proxy
127.0.0.1 - - GET https://example.com 200 1256
```

PR #353, opened 2022-06-15
Add userspace allocator

vinc:
WIP

PR #544, opened 2023-11-18
Add basic userspace allocator

vinc:
This PR continues the work started a year and a half ago here: https://github.com/vinc/moros/pull/353


vinc:
```
~
> print "Hello, World!"
Hello, World!

~
> hello
Hello, World!

~
> hello "Alice"
DEBUG: syscall::alloc(size=7, align=1) -> ptr=0x444446d3a010
DEBUG: syscall::alloc(size=14, align=1) -> ptr=0x444446d3a020
DEBUG: syscall::free(ptr=0x444446d3a010, size=7, align=1)
Hello, !
DEBUG: syscall::free(ptr=0x444446d3a020, size=14, align=1)
```

vinc:
Demo:

```
~
> print "Hello, World!"
Hello, World!

~
> hello
Hello, World!

~
> hello "Alice"
Alice
DEBUG: syscall::alloc(size=7, align=1) -> ptr=0x444446d3a010
DEBUG: syscall::alloc(size=14, align=1) -> ptr=0x444446d3a020
DEBUG: syscall::free(ptr=0x444446d3a010, size=7, align=1)
Hello, Alice!
DEBUG: syscall::free(ptr=0x444446d3a020, size=14, align=1)
```

Not everything work yet, for example if we don't print the `Alice` arg first, the string concatenation doesn't work and `Alice` don't get printed the second time, without causing any error.

vinc:
If we write the first `Alice` in the demo above to `stdnull` instead of `stdout` the demo still work. But it's not simply the conversion of the str into bytes and then the split into ptr and len that happen before the actual syscall.

vinc:
Some issues remain but we'll merge this to implement what is needed for a basic userspace shell.

PR #557, opened 2023-12-06
Using bigger instruction port to get more throughput on data register ATA read/write

LukeHale59:
Switched the data register in ata.rs to u32 instead of u16. Using a bigger instruction port on the data register gives you more throughput to disk. In my testing, it was about 15% faster for writes and 20% faster for reads. @vinc

### DIFF
--- a/src/sys/ata.rs
+++ b/src/sys/ata.rs
@@ -22,7 +22,7 @@ enum Command {
 }
 
 enum IdentifyResponse {
-    Ata([u16; 256]),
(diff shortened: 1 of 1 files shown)


vinc:
Sorry about the delay in answering this PR, I've been wondering about it because it seems that on the ATA spec this register is 16 bit max. I'm using MOROS on real hardware with real HDDs so I can't merge it.

I'll probably implement DMA transfers instead of PIO to speed things up instead of optimizing the current driver.

PR #556, opened 2023-12-01
Improve Lisp with Advent of Code 2023 🎄

vinc:
This year again I'll do the easiest parts of the [Advent of Code](https://adventofcode.com/) in Lisp to improve the language with practical examples.

- [x] Add `\b` and `\e` escape sequences
- [x] Add `dict` type (implemented with `BTreeMap`)
- [x] Add `empty?` function to test empty strings
- [x] Add `reject` function as the inverse of `filter`
- [x] Add `put` function to insert an element into a `dict`, `list`, and `string`
- [x] Add `push` function to insert an element at the end of a `list` or `string`
- [x] Add `host` function to resolve hostnames into IP addresses
- [x] Rename `system` function to `shell` (aliased to `sh`)
- [x] Rename `nth` function to `get` and make it work with `dict`
- [x] Load `(load "/lib/lisp/core.lsp")` in REPL automatically
- [x] Test for truthiness in conditions of `if` and `while`
- [x] Update doc

PR #560, opened 2023-12-17
Downgrade toolchain from 2023-12-01 to 2023-10-01

vinc:
Upgrading the rust toolchain from 2023-05-01 to 2023-12-01 (#559) worked but after merging the lisp PR (#556) the `halt` command would reboot the system (probably after a triple fault).

Downgrading to 2023-11-01 fixed the issue on halting with a serial console but not with a vga console where the command would halt forever.

This issue is not new. It could be due to a random userspace and kernelspace memory issue.


vinc:
Reverting the toolchain to 2023-05-01 fix the issue.

vinc:
The 2023-10-01 toolchain seem to work for both serial and vga consoles.

vinc:
The PR #561 can run correctly with the 2023-12-01 toolchain from the `trunk` branch so we might need this PR.

vinc:
Fixed by #573

PR #566, opened 2024-01-06
Add ANSI OSC color palette support

vinc:
Now that we can print escape sequence from the shell (#558) we can easily add support for the `ESC ] P n rr gg bb` sequence to set a color palette compatible with Linux instead of using a kernel function that is not available from userspace.

For example `print "\e]P0282828"` would change the background color to a dark gray.

We might want to terminate the sequence with `ESC \` to be compatible with ECMA-48.

- [x] Refactor VGA palette code to set the colors one by one
- [x] Add OSC dispatch to handle ANSI color palette sequence
- [x] Add warning macro and deprecate `vga set palette <csv>` command
- [x] Add shell scripts to set the palettes
- [x] Add reset palette sequence

References:
- https://man7.org/linux/man-pages/man4/console_codes.4.html
- https://sw.kovidgoyal.net/kitty/graphics-protocol/

### DIFF
--- a/dsk/ini/palettes/gruvbox-dark.sh
+++ b/dsk/ini/palettes/gruvbox-dark.sh
@@ -1,0 +1,16 @@
+print "\e]P0282828\e[1A" # Black
+print "\e]P1458588\e[1A" # Blue
+print "\e]P298971A\e[1A" # Green
+print "\e]P3689D6A\e[1A" # Cyan
+print "\e]P4CC241D\e[1A" # Red
+print "\e]P5B16286\e[1A" # Magenta
+print "\e]P6D79921\e[1A" # Brown (Dark Yellow)
+print "\e]P7EBDBB2\e[1A" # Light Gray
+print "\e]P8A89984\e[1A" # Dark Gray (Gray)
+print "\e]P983a598\e[1A" # Light Blue
+print "\e]PAB8BB26\e[1A" # Light Green
+print "\e]PB8EC07C\e[1A" # Light Cyan
+print "\e]PCFB4934\e[1A" # Light Red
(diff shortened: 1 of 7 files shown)


vinc:
Here is the [gruvbox](https://github.com/morhetz/gruvbox/) dark palette for example:

```sh
print "\e]P0282828\e[1A" # Black
print "\e]P1458588\e[1A" # Blue
print "\e]P298971A\e[1A" # Green
print "\e]P3689D6A\e[1A" # Cyan
print "\e]P4CC241D\e[1A" # Red
print "\e]P5B16286\e[1A" # Magenta
print "\e]P6D79921\e[1A" # Brown (Dark Yellow)
print "\e]P7EBDBB2\e[1A" # Light Gray
print "\e]P8A89984\e[1A" # Dark Gray (Gray)
print "\e]P983a598\e[1A" # Light Blue
print "\e]PAB8BB26\e[1A" # Light Green
print "\e]PB8EC07C\e[1A" # Light Cyan
print "\e]PCFB4934\e[1A" # Light Red
print "\e]PDD3869B\e[1A" # Pink (Light Magenta)
print "\e]PEFABD2F\e[1A" # Yellow (Light Yellow)
print "\e]PFFBF1C7\e[1A" # White
```

At the moment `print` doesn't support a `-n` argument to avoid printing a new line so we move the cursor one line up each time we call the command to cancel the trailing `\n`.

vinc:
Add sequence to reset the color palette:  `print "\e]R"`

vinc:
There was an issue with the first `\e[1A` if the screen was just printed because the cursor would go up past the edge, freezing everything. This commit add checks to avoid that.

PR #567, opened 2024-01-08
Refactor everything to stay below 80 chars per line

vinc:
This PR is a massive refactoring to restrict the source code to 80 chars per line because that's what the screen is capable of showing in VGA text mode. This will allow reading and editing the source code of MOROS from inside MOROS more efficiently.


vinc:
The code was modified automatically by `cargo fmt` first and then by hand.

PR #570, opened 2024-01-24
Add hashbang support to execute scripts

vinc:
The files starting with `[0x7F, b'E', b'L', b'F']` are identified as ELF binaries and those starting with `[0x7F, b'B', b'I', b'N']` are identified as flat binaries.

With this PR the files starting with a hashbang `[b'#', b'!']` will be identified as scripts by the shell and the interpreter following the shebang will be used to run them with the rest of the command line arguments.

See https://en.wikipedia.org/wiki/Shebang_%28Unix%29

### DIFF
--- a/src/usr/shell.rs
+++ b/src/usr/shell.rs
@@ -585,7 +585,7 @@ fn spawn(path: &str, args: &[&str], config: &mut Config) -> Result<(), ExitCode>
         if contents.starts_with("#!") {
             if let Some(line) = contents.lines().next() {
                 let mut new_args = Vec::with_capacity(args.len() + 1);
-                new_args.push(&line[2..]);
+                new_args.push(line[2..].trim());
                 new_args.push(path);
                 new_args.extend(&args[1..]);
                 return dispatch(&new_args, config);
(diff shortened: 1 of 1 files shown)


vinc:
Shell script example:

```
~
> hello
Hello, World!

~
> read /bin/hello
#!shell

print "Hello, World!"
```

Lisp script example:

```
~
> dhcp
ip:  10.0.2.15/24
gw:  10.0.2.2
dns: 10.0.2.3

~
> ntp time.cloudflare.com
1706129474

~
> read /bin/ntp
#!lisp

(load "/lib/lisp/core.lsp")

(var addr (or (host (head args)) (head args)))
(var port 123)
(var socket (socket/connect "udp" addr port))

(var req (map (fun (i) (if (eq? i 0) 0x33 0)) (range 0 48)))
(file/write socket req)
(var res (file/read socket 48))

(var buf (slice res 40 4))
(var time (- (bin->num (concat '(0 0 0 0) buf) "int") 2208988800))
(print time)
```

And this is the main reason why I chose the symbol `#` for comments in this lisp dialect to be aligned with the shell.

vinc:
This last commit allows white spaces between the hashbang and the interpreter: `#! lisp`

PR #576, opened 2024-02-11
Skip ATA drive selection when not needed

vinc:
We don't need to do the operations to select the drive when it has not changed between operations.


vinc:
Here is a disk benchmark in lisp:

```janet
(load "/lib/lisp/core.lsp")

(var path "/tmp/bench.tmp")

(var buf '(0))
(var i 0)
(var n1 (str->num (first args)))
(print (str "Creating a " (^ 2 n1) " bytes buffer"))
(while (< i n1) (do
  (set buf (concat buf buf))
  (set i (+ i 1))))

(var i 0)
(var n2 (str->num (second args)))
(print (str "Writting buffer to file " n2 " time(s)"))
(var t1 (realtime))
(while (< i n2) (do
  (write-binary path buf)
  (set i (+ i 1))))
(var t2 (realtime))
(var t (- t2 t1))

(print "")
(print (str "Duration:    " t " s"))
(print (str "Performance: " (/ (/ (* (^ 2 n1) n2) 1024) t) " KB/s"))
```

vinc:
Here is a few runs before the change:

```
~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    4.790879487991333 s
Performance: 133.58716319293853 KB/s

~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    4.741887092590332 s
Performance: 134.96736373163827 KB/s

~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    5.112830638885498 s
Performance: 125.17527866706496 KB/s

~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    4.918012380599976 s
Performance: 130.13387329495148 KB/s
```

And here is after the change:

```
~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    3.803030014038086 s
Performance: 168.28686537775786 KB/s

~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    3.593909502029419 s
Performance: 178.0790528082588 KB/s

~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    3.6830482482910156 s
Performance: 173.76910560348176 KB/s

~
> lisp bench-disk.lsp 16 10
Creating a 65536 bytes buffer
Writting buffer to file 10 time(s)

Duration:    3.854869842529297 s
Performance: 166.0237637440118 KB/s
```

Still slow but it's a 30% improvement on QEMU

PR #579, opened 2024-02-13
Add time synchronisation with NTP

vinc:
A new `ntp` program will now be available to get the time from NTP servers and update the RTC of the computer. This is the first program written in Lisp using the new hashbang system for scripts (#570).

ntp


vinc:
The default time server can be configured in `/ini/ntp`.

Currently it is set to `time.cloudflare.com`. I had an issue with a lot of servers from `pool.ntp.org` that I fixed by changing a flag in the first byte of the network packet, but another less frequent one remain it seems.

The NTP client I wrote is pretty trivial but it does the job for a small personal system.

Issue #426, opened 2022-10-27
Error: MFS is not mounted to '/'


yonas:
I tried running MOROS in qemu and got this error:

```
MFS is not mounted to '/'
Running console in diskless mode
```

Cargo: `1.63.0-nightly (8d42b0e87 2022-06-17)`
OS: `Ubuntu 22.10`
QEMU: `7.0.0 (Debian 1:7.0+dfsg-7ubuntu2)`

yonas:
I `did not` run these two commands, since rust is already installed:

```
$ rustup install nightly
$ rustup default nightly
```

```
$ cargo install bootimage

info: syncing channel updates for 'nightly-2022-06-21-x86_64-unknown-linux-gnu'
info: latest update on 2022-06-21, rust version 1.63.0-nightly (5750a6aa2 2022-06-20)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'llvm-tools-preview'
info: downloading component 'rust-docs'
info: downloading component 'rust-src'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'llvm-tools-preview'
info: installing component 'rust-docs'
info: installing component 'rust-src'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
```

vinc:
Hi @yonas, thanks for trying MOROS!

In diskless mode you can run `install` to install the system as described in the README.

I've been thinking a few times about printing that suggestion after the warning because that's what most users will want to run here.

vinc:
Hi @yonas, thanks for trying MOROS!

In diskless mode you can run `install` to install the system as described in the README.

I've been thinking a few times about printing that suggestion after the warning because that's what most user will want to run here.

yonas:
@vinc Thanks, that worked! :partying_face: 

> I've been thinking a few times about printing that suggestion after the warning because that's what most user will want to run here.

I agree, this would be very helpful.

PR #595, opened 2024-03-13
Adds a helpful println when MFS is not mounted

kidharb:
First time users get stuck after booting disk.img for the first time (as did I). 
This PR adds a helpful debug line which tells the user to run 'install'.
Addresses issue #426

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,6 +33,7 @@ fn user_boot() {
             println!("Could not find '{}'", script);
         } else {
             println!("MFS is not mounted to '/'");
+            println!("Please run 'install' to setup moros");
         }
         println!("Running console in diskless mode");
         usr::shell::main(&["shell"]).ok();
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,7 +32,6 @@ fn user_boot() {
         if sys::fs::is_mounted() {
             println!("Could not find '{}'", script);
         } else {
-            println!("MFS is not mounted to '/'");
             println!("MFS not found, run 'install' to setup the system");
         }
         println!("Running console in diskless mode");
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,7 +33,7 @@ fn user_boot() {
             println!("Could not find '{}'", script);
         } else {
             println!("MFS is not mounted to '/'");
-            println!("Please run 'install' to setup moros");
+            println!("MFS not found, run 'install' to setup the system");
         }
         println!("Running console in diskless mode");
         usr::shell::main(&["shell"]).ok();


src/main.rs
```diff
@@ -33,6 +33,7 @@ fn user_boot() {
             println!("Could not find '{}'", script);
         } else {
             println!("MFS is not mounted to '/'");
+            println!("Please run 'install' to setup moros");
```
vinc:
```suggestion
            println!("MFS not found, run 'install' to setup the system);
```

vinc (changes requested):

src/main.rs
kidharb: Happy with this. I also tested and it is as expected. (You missed a " after system :-P)

vinc:
Thanks a lot for your feedback!

We can merge the output with the previous line to make it less verbose but still helpful.

I'll merge your PR and create another one to add the option to setup the file system in memory from the `install` command instead of having to run `memory format` first as explained in the manual. It's a usecase I have a lot when I don't want to write to disk or when no disks have been found and that's why I wasn't sure about simply suggesting to run `install`.

I should also link the manual from the README!

kidharb:
Merging the debug would be better. 

I think it would be better to cater for the majority of the people who join the project, don't read the manual (I am guilty) but want to get up and running by following the build and run instructions.

Setting up the filesystem is a use case I want as well, saves me from removing the disk every time I want to test the installation process.

src/main.rs
```diff
@@ -33,6 +33,7 @@ fn user_boot() {
             println!("Could not find '{}'", script);
         } else {
             println!("MFS is not mounted to '/'");
+            println!("MFS not found, run 'install' to setup the system");
```
vinc:
```suggestion
            println!("MFS not found, run 'install' to setup the system");
```

vinc (changes requested):

vinc:
I agree about the manual, it's good to have one but the system should be usable without reading it.

vinc (approved):
Perfect, thank you!

kidharb:
Aside, I would like to chat on what is priority to work on? 
But I also do not want it to become management overhead for you. 

This project is exactly what I was looking for 
Rust --> Check
Small OS to learn stuff --> Check

I do also want to learn about TLS v1.3 and would like to create a utility called `mossl` which will be slightly similar to `openssl` and perform a SSL handshake with an endpoint.
Can I go ahead and work on that?

vinc:
MOROS is still experimental at this stage and will change a lot before it reaches version 1.0 so I try to discourage other devs to work on big features yet to avoid wasting their time. But if you want to use it as a base to try an idea feel free to fork it and hack it, a few people have done that in the past and it's pretty cool!

My main goal is to build a simple operating system that can fit in one standard human brain. I hope to encourage OS diversity and see more people having their own hobby OS. Essentially answering the question of what to do after finishing an osdev tutorial with MOROS as an example.

I reworked the network stack last year to add some syscalls and device files for the sockets (#512) but I still have memory issues preventing me to run complex programs fully in userspace. At the moment I also completely bypassed the lack of TLS support by creating a HTTP/0.9 proxy (https://github.com/vinc/fetch) that I run on another computer.

Feel free to reach me by email (v@vinc.cc) if you want!

PR #596, opened 2024-03-13
Improve setup process

vinc:
This PR is a continuation of #595 by @kidharb to improve the setup process

- [x] Use `warning` macro when MFS is not present
- [x] Use `error` macro when `/ini/boot.sh` is not present
- [x] Remove diskless mode message
- [x] Add option to setup the system on a RAM disk from the `install` command

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,7 +5,9 @@ extern crate alloc;
 
 use bootloader::{entry_point, BootInfo};
 use core::panic::PanicInfo;
-use moros::{debug, hlt_loop, print, println, sys, usr};
+use moros::{
+    debug, error, warning, hlt_loop, eprint, eprintln, print, println, sys, usr
+};
 
 entry_point!(main);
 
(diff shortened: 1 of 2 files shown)


vinc:
Here's the warning when MFS is not present:

install

And here's the error when `/ini/boot.sh` is not present:

init

vinc:
Users will now be able to setup the system on a RAM disk directly from the `install` command instead of having to run `memory format` first:

```
/
> install
Welcome to MOROS v0.10.2 installation program!

Proceed? [y/N] y

Listing disks ...
Path            Name (Size)
/dev/ata/0/0    QEMU HARDDISK QM00001 (32 MB)
/dev/mem        RAM DISK

Formatting disk ...
Enter path of disk to format:
```

kidharb:
> And here's the error when `/ini/boot.sh` is not present:
> 
> init

How do you recover in the case of missing `/ini/boot.sh`? 

Do we need to create it and add?
```
user login
shell
```

vinc:
> How do you recover in the case of missing `/ini/boot.sh`?
> 
> Do we need to create it and add?
> 
> ```
> user login
> shell
> ```

The quickest way would be to just run `install` again that will recreate any missing files on the mounted file system. That's how I update the system whenever I change something: I delete the files and run `install` again.

But you can also create your own boot script indeed, for example sometimes I just have `lisp` to boot into the Lisp interpreter, and on my test computers I add `env TZ 3600` in winter to set the correct timezone.

Here's the current version of the manual: http://moros.cc/manual.html

PR #597, opened 2024-03-15
Update unit conversions

vinc:
Until now we were using an algorithm similar to `ls -lh` and `ls -l --si` from Linux to convert number to a readable format, either binary (1K == 1024) or decimal (1K == 1000), but some numbers like 1120 and 1160 didn't seem to be correctly rounded. So now we use a simpler algorithm with correct results according to our tests. For example 1160 should be 1.1K and 1.2K respectively.

### DIFF
--- a/src/api/unit.rs
+++ b/src/api/unit.rs
@@ -9,39 +9,53 @@ pub enum SizeUnit {
 }
 
 impl SizeUnit {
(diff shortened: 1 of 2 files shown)


vinc:
We might want to use uppercase `K` for binary and lowercase `k` for decimal.

PR #602, opened 2024-03-21
Improve RNG

vinc:
The RNG module uses HC-128 to generate bytes for `/dev/random`. It is seeded by either the RDRAND instruction if available or a combination of uptime and realtime clocks.

Initially we were seeding a new RNG each time a random byte was requested, which is not efficient and it was causing issues when RDRAND was not available because each requests between the same ticks would return the same bytes.

With this PR we'll seed the RNG during boot so that consecutive requests will generate a different random byte, and we'll use SHA-256 when RDRAND is not available to mix different entropy sources. For now it will still only be the uptime and realtime clocks but in the future we can reseed the RNG regularly and use more sources like keyboard, hard drive, network card... Another interesting thing is that we can disable this init process to use the default zeroed seed and have a fully deterministic RNG if needed.

Reading `/dev/random` will now have the same performance regardless of how we seeded the RNG which was not the case before. We also improved how the RNG is used when reading the device file if a large buffer is requested.

In the process of writing this PR we also ended up fuzzing the VGA driver and discovering a few bugs causing a freeze after some amount of time printing random bytes to screen.


vinc:
Reading `/dev/random` will freeze the computer after a while but I'm not quite sure why. I tried a lot of things like suppressing interrupts or using atomics instead of the mutex without success.

vinc:
Turns out the issue is not really the RNG, it's simply that reading `/dev/random` to the console is acting as a fuzzer for the system, notably the VGA driver.

vinc:
random3

I have a wall of computers reading `/dev/random` to see if I can find more bugs, it's fun!

PR #606, opened 2024-03-25
Add virtio-net driver

vinc:
- https://ozlabs.org/~rusty/virtio-spec/virtio-0.9.5.pdf
- https://docs.oasis-open.org/virtio/virtio/v1.0/virtio-v1.0.html


vinc:
Getting the MAC address:

virtio-net

Now the next step is to build the virtqueue which is a bit harder due to memory requirements...

PR #620, opened 2024-05-25
Add device creation from write command

vinc:
This will simplify the creation of device files from userspace with the following command: `write <path> --dev <name>`. It will allow again manual (or scripted) installation of the system.


vinc:
```
~
> write clock --dev rtc

~
> read clock
2024-05-25 17:47:50

~
> delete clock

~
> write clock --dev realtime

~
> read clock
1716659284.355946
```

PR #627, opened 2024-05-29
Add scrollback buffer support

vinc:
Using the `PAGE_UP` and `PAGE_DOWN` keys in the console, especially in the shell, will now scroll the screen up or down and display what was previously on the screen before the current lines replaced it.


vinc:
The idea is simple: we have a text buffer in the VGA driver to print characters to the 80x25 screen, the scrollback buffer will be the same thing but larger (80x2500 for example), and every time we change something in the text buffer we also change it in the scrollback buffer.

The scrollback buffer will have a moving 80x25 window representing the current text buffer and the rest will contain what was on the screen in the past. We will need a circular buffer to overwrite the oldest lines with the new ones.

The implementation is complicated by the line editing capabilities of the console. We will probably need the concept of alternate buffer to do line editing. Currently it works for printing the kernel logs at boot but some of the screen is missing when we arrive in the shell. The circular buffer logic has not been implemented yet. Right now we are listening for the page up and down keys in the prompt, but we should do that elsewhere.

vinc:
Now we need an alternate buffer where scrollback is disabled for the text editor/viewer.

vinc:
If clearing the screen also clear the scroll back buffer then we don't need alternate screen buffer support for now.

PR #630, opened 2024-05-31
Add file search to find command

vinc:
This PR implement file search in `find` and rename it `--file <pattern>` instead of `--name <pattern>` where `<pattern>` allow globbing like the shell. In `--line <pattern>` it is a standard regular expression.

By default `find` will be equivalent to `find "$DIR" --file "*"` instead of `find --help`.


vinc:
find-file

vinc:
It should be possible to do `find / --file "*.lsp" --line "print"` to recursively search a line pattern inside files matching a file pattern in the path.

It could also be possible to search in more than one path.

vinc:
The `find.png` screenshot in the doc is not valid anymore because the regex syntax must be escaped since #558 and we introduced a new file in #618 that matches the line pattern in the example.

vinc:
New screenshot for the documentation:

find

vinc:
We can now use both `--file` and `--line` to filter in which files we want to search the given lines:

```
~
> find /tmp --line "list"
/tmp/alice.txt
63: was no one to listen to her, still it was good practice to say it over)
71: there WAS no one listening, this time, as it didn't sound at all the

/tmp/lisp/doc.lsp
9:     (if (nil? s) "" (str " " (if (list? s) (str/join s " ") s))) # args

~
> find /tmp --file "*.lsp" --line "list"
/tmp/lisp/doc.lsp
9:     (if (nil? s) "" (str " " (if (list? s) (str/join s " ") s))) # args
```

And the code has been refactored to be DRY.

vinc:
Running `find lisp` inside `/tmp` would print files with a relative path starting with `lisp/` but running `find` inside `/tmp/lisp` would print files with an absolute path starting with `/tmp/lisp/`. This inconsistency is now fixed:

```
~
> find /tmp/lisp
/tmp/lisp/colors.lsp
/tmp/lisp/doc.lsp
/tmp/lisp/factorial.lsp
/tmp/lisp/fibonacci.lsp
/tmp/lisp/geotime.lsp
/tmp/lisp/pi.lsp
/tmp/lisp/sum.lsp
/tmp/lisp/tak.lsp

~
> /tmp/lisp

/tmp/lisp
> find
colors.lsp
doc.lsp
factorial.lsp
fibonacci.lsp
geotime.lsp
pi.lsp
sum.lsp
tak.lsp
```

PR #635, opened 2024-06-04
Fix final line ending

vinc:
The previous PR #632 introduced an issue on empty files, because `"".split('\n')` would produce `[""]` whereas `"".lines()` would produce `[]` resulting in a panic on the editor and the viewer.

We can add back an empty line if a file contains no lines, but it's also the occasion to deal with something that could be improved: in POSIX the last line of a file must end with a newline, and this make sense so MOROS follows this rule. But the editor would explicitely show an empty line at the end of every file to save this trailing newline chars. And when editing a file you'd have to remember to always leave an empty line at the end.

In general it seems better to make that final newline implicit and avoid displaying an empty line at the end of every files.

The only caveat is that on some occasions we might want to produce a file without a trailing newline. This will now be impossible, we'll see if it's needed in the future and revisit that decision then. It could just be an option in the editor.

### DIFF
--- a/src/usr/edit.rs
+++ b/src/usr/edit.rs
@@ -56,6 +56,9 @@ impl Editor {
                 for line in contents.lines() {
                     lines.push(line.into());
                 }
+                if lines.is_empty() {
+                    lines.push(String::new());
+                }
             }
             Err(_) => {
                 lines.push(String::new());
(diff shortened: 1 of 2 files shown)


vinc:
Interesting read: https://stackoverflow.com/a/729795

PR #637, opened 2024-06-06
Fix block bitmap utilisation

vinc:
I found another issue after fixing #636. We were using only 1/8 of every block bitmap to mark allocated blocks resulting in an exception when trying to use more than one eight of the disk:

```
~
> dhcp
[23.345445] NET IP 10.0.2.15/24
[23.348445] NET GW 10.0.2.2
[23.352444] NET DNS 10.0.2.3

~
> dsk u -b
size: 8.0M
used: 108K
free: 7.9M

~
> time http 10.0.2.2:8000/1000kb.txt => 1
Executed 'http 10.0.2.2:8000/1000kb.txt' in 14.685895s

~
> dsk u -b
size: 8.0M
used: 1.1M
free: 6.9M

~
> time http 10.0.2.2:8000/1000kb.txt => 2
DEBUG: panicked at src/sys/fs/bitmap_block.rs:25:13:
index out of bounds: the len is 512 but the index is 512
qemu-system-x86_64: terminating on signal 2
```

PR #638, opened 2024-06-06
Fix block bitmap area calculation

vinc:
The formula used to calculate the size of the block bitmap area was not correct, resulting in blocks that cannot be allocated. This issue surfaced after #637 when using about half of the disk.

One thing to note is that this PR will likely extend the size of the block bitmap area by one or more blocks so we must increase the version number of MFS to use the old formula on existing disks.

### DIFF
--- a/doc/filesystem.md
+++ b/doc/filesystem.md
@@ -10,9 +10,9 @@ A hard drive is separated in blocks of 512 bytes, grouped into 4 areas:
     +------------+
     | Superblock | (2 blocks)
     +------------+
-    | Bitmap     | (n / (8 * 512) blocks)
+    | Bitmap     | (n blocks)
     +------------+
-    | Data       | (n blocks)
+    | Data       | (n * 512 * 8 blocks)
     +------------+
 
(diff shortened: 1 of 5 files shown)


vinc:
This PR also fixes out of space errors when we used the whole disk.

vinc:
And now when we are out of disk we get the correct numbers:

```
~
> disk usage
size: 8376320
used: 8376320
free:       0
```

The system doesn't crash, there's just no writes to disk possible.

vinc:
To test the disk I generated a number of files with precise sizes on the host like this 100 KB file:

```
$ dd if=/dev/zero of=100 bs=1024 count=100
```

Then I launched a HTTP server in the dir on the host:

```
$ python -m http.server
```

And I downloaded the file from MOROS:

```
> time http 10.0.2.2:8000/100 => 1
```

vinc:
disk-4

vinc:
This is a breaking change that is very close to the summer solstice release date of MOROS so we'll merge it after the release to have 3 months to test it before the next.

PR #653, opened 2024-09-12
Add VgaFont device

vinc:
A new device in `/dev/vga/font` will allow setting the VGA font from userspace. After that we can deprecate the whole `vga` command and remove it in the next version of MOROS.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -288,8 +288,8 @@ You can also set the `TZ` environment variable to use your preferred timezone:
 Add `env TZ 7200` to `/ini/boot.sh` before `shell` to save the timezone:
 
     > read /ini/boot.sh
-    vga set font /ini/fonts/zap-light-8x16.psf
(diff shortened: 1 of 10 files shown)


vinc:
There's an issue when doing `read /ini/fonts/zap-light-8x16.psf => /dev/vga/font` because `read` try to convert the file to unicode before printing it to the screen:

vga-font-bug

And `copy` doesn't work for device files.

vinc:
This change will allow writing to devices.

We could also replace `copy` with a lisp script doing the following at its core:

```janet
(write-binary (second args) (read-binary (first args)))
```

vinc:
This change allow reading binary files, so we can now do:

    > read /ini/fonts/zap-light-8x16.psf => /dev/vga/font

In addition to:

    > copy /ini/fonts/zap-light-8x16.psf /dev/vga/font

I'm not sure that the latter is the right thing because we could also expect the source to replace the destination instead of writing to it when it's a device file.

PR #654, opened 2024-09-13
Replace base64 command with encode and decode

vinc:
The `base64` command was only used as a library by the `user` command so its content can be moved to the API and we can introduce two new commands, `encode` and `decode`, that will also use the API to encode and decode files (instead of arguments).


vinc:
base64

PR #655, opened 2024-09-14
Add filetype syscall

vinc:
This will add a new `KIND` syscall to get the `FileType` of a file handle. It will be used to detect IO redirections and allow some commands to disable colors when their output is redirected to a file.


vinc:
One limitation of this implementation is that it won't detect redirections to device files, only regular files because `FileType` doesn't distinguish between the console device and other kind of devices.

vinc:
The hash command has been modified to disable colors when redirecting to file.

PR #656, opened 2024-09-14
Update hash command

vinc:
The `hash` command can make use of #655 to display more abbreviated hashes on the console and the full hashes when redirected to a file. The command line options have to be modified to allow this change.


vinc:
hash

With this modification we are choosing to maximize the default usefulness of a program even if it means different outputs depending on how the program is used. The alternative was to disable `--short` and `--color` by default to have the same behavior all the time.

PR #659, opened 2024-09-16
Add search to editor

vinc:
This PR finally add a search functionality to the editor! It will also be the first step to adding a command mode.

edit-find


vinc:
Use `CTRL + F` to open the search mode and find the first result of the query then `CTRL + N` to find the next result.

edit-find-next

vinc:
We can use the prompt library to have line editing and history out of the box while keeping the code DRY.

PR #660, opened 2024-09-17
Add a line mode to the editor

vinc:
This PR is a test to see how useful it would be to have a tiny subset of ed/vi commands in the editor.

### DIFF
--- a/src/api/regex.rs
+++ b/src/api/regex.rs
@@ -98,6 +98,19 @@ impl Regex {
         }
         res
     }
+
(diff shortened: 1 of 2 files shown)


vinc:
Testing a substitute command:

edit-line

vinc:
Now with regex support for the substitute command:

edit-line-regex

(note that the `g` flag is gone, we need to also add `Regex#replace_all` for it)

vinc:
The `--command` transform the editor into a kind of "stream editor" like sed and it would be easy to do `alias sed "editor --command"`.

edit-line-arg

PR #662, opened 2024-09-19
Add chess puzzle

vinc:
Add a `puzzle <path>` command to the `chess` program that works like `load <path>` but will randomly load a puzzle from a list of FEN positions contained in an EPD file:

chess-puzzle


vinc:
The puzzles come from here: https://wtharvey.com/m8n2.txt

The EPD file was created like this:

```
$ curl -sSL https://wtharvey.com/m8n2.txt | grep " - [01]" | dos2unix > dsk/tmp/chess/mi2.epd
```

vinc:
If the whole puzzle collection is compiled in the system image its size goes over the 2 MB limit so there's two options for the future: either I make a breaking change to reserve 4 MB at the beginning of the disk for the kernel instead of 2 MB, or I update the installer to download non-essential files from the internet.

There's already a breaking change for the next version of MOROS that fixes the way disk block were allocated so it's a good time for that, and MOROS have been just short of the 2 MB limit for a very long time, but having a light weight kernel and downloading files is a more standard way to do things.

I'll leave this decision out of this PR. I'll probably need to introduce file compression either way to be more storage efficient.

disanv03:
1. Nf6+, gxf6 2.Bxf7# nice puzzle :)

Didn't know about the epd file format interesting. Nice trick to collect all the fen position with grep.

Very pretty terminal board !

PR #663, opened 2024-09-21
Add pkg command

vinc:
This PR will introduce the `pkg install <name>` command to fetch files that are not bundled in the OS image in an effort to reduce the size of the latter.

In order for this to work the host (or any computer on the network) must run the pkg server found in `run/pkg-server.sh` which is just a wrapper around `python -m http.server` to serve the content of the `dsk` dir from this git repo. If this experiment is validated we'll host each version of MOROS on http://pkg.moros.cc but MOROS don't support TLS yet so we can't use HTTPS which make this solution less than ideal.

For now only non essential files found in `/tmp` can be installed this way. For example MOROS Lisp example programs or cell files for the game of life are found there.

I looked into the `zip` crate but it doesn't support `no_std` so a package is just a text file listing the path of the files that should be downloaded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,9 @@ test:
 website:
 	cd www && sh build.sh
 
+pkg:
+	ls -1 dsk/var/pkg | grep -v index.html > dsk/var/pkg/index.html
+
 clean:
 	cargo clean
 	rm -f www/*.html www/images/*.png
--- a/doc/lisp.md
+++ b/doc/lisp.md
@@ -75,11 +75,12 @@ MOROS Lisp is a Lisp-1 dialect inspired by Scheme, Clojure, and Ruby!
 - `regex/match?`
 
 ### File Library
+- `dirname`, `filename`
 - `read`, `write`, `append`
 - `read-binary`, `write-binary`, `append-binary`
 - `read-line`, `read-char`
 - `uptime`, `realtime`
-- `p`, `print`
+- `p`, `print`, `eprint`, `error`
 
 ### Math Library
 - `floor`, `ceil`, `round`
@@ -174,6 +175,7 @@ Would produce the following output:
 ## Changelog
 
 ### Unreleased
+- Add `dirname`, `filename`, `eprint`, and `error` functions
 
 ### 0.7.1 (2024-06-20)
 - Add `floor`, `ceil`, and `round` functions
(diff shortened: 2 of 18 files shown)


vinc:
The latest commits make use of #664 to create parent directories if needed before fetching the files so we don't have to create them before calling `pkg install <name>`.

vinc:
The `http` command will now exit with 0 only if the HTTP response code is below 400, otherwise it will exit with 1.

New `eprint` and `error` functions where added to MOROS Lisp to format error messages in the `pkg` command written in MOROS Lisp the same way than commands written in Rust.

pkg

vinc:
Use the same colors as regular programs written in Rust:

pkg-usage

vinc:
The makefile was updated to automatically build the package index:

pkg-list

vinc:
The `remove` command try first to use the local package file to list the files that need to be deleted and fallback to fetching the latest package file. This ensure that files no longer in the latest package file but present when the package was installed will be correctly deleted.

```
~
> pkg -h
Usage: pkg <cmd>

Commands:
  list             List packages
  install <pkg>    Install package
  remove <pkg>     Remove package

~
> pkg i chess
Fetching '/var/pkg/chess'
Fetching '/tmp/chess/mi2.epd'
Fetching '/tmp/chess/mi3.epd'
Fetching '/tmp/chess/mi4.epd'
Fetching '/tmp/chess/puru.epd'

~
> pkg r chess
Deleting '/tmp/chess/mi2.epd'
Deleting '/tmp/chess/mi3.epd'
Deleting '/tmp/chess/mi4.epd'
Deleting '/tmp/chess/puru.epd'

~
> pkg i chess
Fetching '/var/pkg/chess'
Fetching '/tmp/chess/mi2.epd'
Fetching '/tmp/chess/mi3.epd'
Fetching '/tmp/chess/mi4.epd'
Fetching '/tmp/chess/puru.epd'

~
> del /var/pkg/chess

~
> pkg r chess
Fetching '/var/pkg/chess'
Deleting '/tmp/chess/mi2.epd'
Deleting '/tmp/chess/mi3.epd'
Deleting '/tmp/chess/mi4.epd'
Deleting '/tmp/chess/puru.epd'
```

vinc:
If this experimental package manager is validated I'll have to update the `README` and the `doc` before the next release, and create a new `moros-packages` git repository to host the package files (example: http://download.moros.cc/releases/0.10.4/files/var/pkg/chess)

PR #668, opened 2024-09-26
Fix userspace heap address

vinc:
There was an issue with userspace alloc for a very long time when trying to push a program argument in a string allocated in userspace. The arguments are allocated just below the heap, turns out we needed to start the heap one page above.

Before we had to print an arg first (weird) to be able to push it to the string:

```
> hello alice bob
alice
Hello, alice!
Hello, !
```

Now we don't need to do that anymore:

```
> hello alice bob
alice
Hello, alice!
Hello, bob!
```

But we still need to use `ld` as a linker instead of `lld` to avoid another (weird) bug:

```
> hello.rust-lld alice bob
alice
DEBUG: Could not map Page[4KiB](0x207000) to PhysFrame[4KiB](0x15ba000)
DEBUG: Aleardy mapped to PhysFrame[4KiB](0x407000)
Error: Could not allocate address 0x207790
```

This address is the beginning of the `.bss` segment that is only present when the code need the userspace allocator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,8 @@ user-rust:
 		touch dsk/bin/{}
 	basename -s .rs src/bin/*.rs | xargs -I {} \
 		cargo rustc --no-default-features --features userspace --release --bin {}
+	cargo rustc --no-default-features --features userspace --release --bin hello -- -C linker-flavor=ld
+	cargo rustc --no-default-features --features userspace --release --bin exec -- -C linker-flavor=ld
 	basename -s .rs src/bin/*.rs | xargs -I {} \
 		cp target/x86_64-moros/release/{} dsk/bin/{}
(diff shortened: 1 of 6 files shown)


vinc:
With that command arguments fix we can now make a much more useful user shell!

userspace-shell-2

vinc:
Now that the bug is fixed we can rewrite the `hello` program to print `Hello, <arg>!` for each of its args or `Hello, World!` if none.

vinc:
There's still an error if we use the `format` macro instead of `push_str`:

```
> hello a b
DEBUG: Could not map Page[4KiB](0x402000) to PhysFrame[4KiB](0x15bb000)
DEBUG: Aleardy mapped to PhysFrame[4KiB](0x76000)
Error: Could not allocate address 0x402F20
```

PR #304, opened 2022-01-12
Add geodate program

vinc:
Geodate displays the current local time in a geocentric date format using a natural lunisolar calendar with metric time based on decimal fractions of the mean solar day. See https://github.com/vinc/geodate

The program is rather large with lots of astronomical computations so it's a good test for MOROS.

PR #673, opened 2024-09-28
Add special alloc_error_handler for userspace

vinc:
Userspace programs need their own version of `alloc_error_handler` to avoid using kernel functions to print the error.

This was discovered here: https://github.com/vinc/moros/pull/304#issuecomment-2380636887


vinc:
With this we can recover allocation errors:

```
~
> /bin/geodate -3.3333
Error: Could not allocate address 0x40C300

~
>
```

PR #678, opened 2024-09-30
Fix shell autocomplete command path

vinc:
There was a special case in path autocompletion of the shell for the first arg when it a path to a binary, but it doesn't seem to be useful anymore, because every file can be a script now, so this code can be simplified.

### DIFF
--- a/src/api/fs.rs
+++ b/src/api/fs.rs
@@ -61,6 +61,10 @@ pub fn realpath(pathname: &str) -> String {
     }
 }
 
+pub fn is_absolute_path(path: &str) -> bool {
+    path.starts_with('/') || path.starts_with('~')
+}
+
 pub fn exists(path: &str) -> bool {
     syscall::info(path).is_some()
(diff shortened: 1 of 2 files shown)


vinc:
The condition to autocomplete commands is `i == 0 && !args[i].starts_with('/') && !args[i].starts_with('~')` so we could decide to use something like `(i == 0 && (args[i].starts_with('/') || args[i].starts_with('~'))) || i > 0` if we wanted to only allow running a script or a binary from its path using either `/full/path/to/binary` or `~/home/path/to/binary`.

vinc:
Now the autocompletion rules of path to binary for the first arg and path to resource for the other args are the same, but the path to binary must be an absolute path (starting with `/` or `~`).

vinc:
It is still possible to run a binary or a script, or jump to a directory, from a relative path but they wont be autocompleted to avoid cluttering the list of candidates.

PR #684, opened 2024-10-04
Improve Lisp

vinc:
I'm doing a Leetcode challenge in October with MOROS List to improve the language.
- [x] Fix dict representation when empty
- [x] Add dict support to `contains?`
- [x] Add string support to `head` and `tail`

### DIFF
--- a/src/usr/lisp/eval.rs
+++ b/src/usr/lisp/eval.rs
@@ -47,11 +47,15 @@ fn eval_head_args(
 ) -> Result<Exp, Err> {
     ensure_length_eq!(args, 1);
     match eval(&args[0], env)? {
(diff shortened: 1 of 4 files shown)


vinc:
```
~
> lisp
MOROS Lisp v0.7.0

> (dict)
(dict)

> (dict "foo" 42 "bar" 1337)
(dict "bar" 1337 "foo" 42)

> (contains? (dict "foo" 42) "foo")
true

> (contains? (dict "foo" 42) "bar")
false

> (head "foo") # Equivalent to `(first "foo")` which is derived from `(get "foo" 0)`
"f"

> (tail "foo") # Aliased to `(rest "foo")`
"oo"
```

PR #687, opened 2024-10-11
 Add VGA graphics

vinc:
This PR started as a test to see if we could switch between VGA text mode and graphics mode when the OS is running and after a quick test with the `vga` crate the answer is yes!

This part of the `vga` crate is based on this link: https://www.singlix.com/trdos/archive/vga/Graphics%20in%20pmode.pdf

I used the PDF to do the same thing directly inside the VGA driver in MOROS.

After that I tried to load a simple PPM image created in Gimp (and generated by a LLM) while still using `set_pixel()` from the `vga` crate.

The idea is to end up with a framebuffer and the goal to finally be able to run DOOM.


vinc:
When we try to display this image with the default VGA palette:

forest_320x200_256

We get this:

vga-graphics-14

We see the outlines of the image but the colors are wrong.

vinc:
When we try to display all the colors in the palette we only see 64 colors, the rest is black:

vga-graphics-13

vinc:
If we update the VGA driver to allow setting up the 256 colors palette we get the full image:

vga-graphics-15

We could now use another image format like BMP instead of PPM to join a custom palette to get better colors.

vinc:
Here is a gist I used for importing the default VGA palette in Gimp: https://gist.github.com/TatuArvela/c66de7e0174a54bc6231be85dd4e4e7f#file-vga-256-colors-gpl

vinc:
The indexes of the 16 colors palette is not the same in the VGA driver and in the console so we need to translate them, but we can skip the translation for the 256 colors palette.

The previous images had some wrong colors like the navy blue pixels in the trunks on the right that are now gone:

vga-graphics-16

vinc:
The code is a mess but we can now read BMP images with 256 indexed colors that we can load into the VGA palette to improve the rendering:

vga-graphics-21

vinc:
Now that we can write directly to the VGA framebuffer we don't need the `vga` crate anymore, and loading an image at once is much faster than using `mode.set_pixel(x, y, c)` which was to be expected.

vinc:
The `image` command was renamed `render` to be a verb. It could also have been `display` or `draw`.

The `read` command will dispatch `*.bmp` files to the `render` command like it does for directories to the `list` command.

So we can do `read example.bmp` or `render example.bmp` directly.

vinc:
We can now do `render *.bmp` and use the arrow keys to cycle through the images.

Quit with `ctrl+c` or `ctrl+q`.

vinc:
The next step will be adding the necessary device files and operations to be able to set the mode from userspace and backup/restore the 256-colors palette and the font. The `vga` command will also have to be removed to allow refactoring that part of the VGA driver, and the driver will probably better be split into multiple files inside a module, but this PR is already too big to do all of that.

PR #690, opened 2024-10-18
Create memory module

vinc:
- [x] Take unmapped and virtual memory into account when counting total memory
- [x] Enable interrupts except keyboard during memory init to improve boot time accuracy
- [x] Create `mem` module
- [x] Move `allocator` inside `mem` module split into `heap` and `phys`
- [x] Update `memory usage` command to report whole RAM usage


vinc:
Allowing interrupt during kernel heap initialization correctly count elapsed time during the allocation, which can be slow with large memory. No adverse effects have been found yet.

vinc:
Typing on the keyboard during memory init generates interrupts that crash the module. This can be fixed by delaying keyboard init to after memory init.

vinc:
Except that loading the keyword module after memory initialization breaks it. Masking the keyword interrupt works though.

vinc:
The `memory usage` command will now report whole RAM usage instead of only the kernel heap.

PR #691, opened 2024-10-20
Create clock module

vinc:
Before:

- `src/sys/clock.rs`
- `src/sys/cmos.rs`
- `src/sys/time.rs`

After:

- `src/sys/clk/mod.rs`
- `src/sys/clk/boot.rs`
- `src/sys/clk/epoch.rs`
- `src/sys/clk/rtc.rs`
- `src/sys/clk/cmos.rs`
- `src/sys/clk/sync.rs`
- `src/sys/clk/timer.rs`


vinc:
This last commit is a breaking change for the API and the device files:

clk

- The clock at `/dev/clk/boot` (previously at `/dev/clk/uptime`) returns the number of seconds since boot
- The clock at `/dev/clk/epoch` (previously at `/dev/clk/realtime`) returns the number of seconds since epoch
- The clock at `/dev/clk/rtc` (previously at `/dev/rtc`) returns the value of the RTC

vinc:
This last commit prevent panicking when writing to some device files where writing was not implemented like `copy something /dev/clk/epoch`, instead the operation will return an error.

PR #693, opened 2024-10-22
Add speaker device

vinc:
This PR will add `/dev/speaker` allowing the `beep` command to interact with the PC Speaker from usermode.

Writing a (big endian) bytes representation of a 64 bit float to this device will start to play a sound at the given frequency or stop it if the frequency is zero.


vinc:
If we write the frequency to the device using a string instead of a float we can use the shell more easily to play sound:

speaker

And the `beep` command could be a simple script that uses `print` to write the frequency to the device, then `sleep` to wait for the sound to play, then `print` again to stop the sound:

```
print $1 => /dev/speaker
sleep $2
print 0 => /dev/speaker
```

vinc:
The `beep` command can now become a regular userspace program.

PR #700, opened 2024-11-06
Add network config devices

vinc:
- [x] Add `net-gw` device
- [x] Add `net-ip` device
- [x] Add `net-mac` device
- [x] Add `net-usage` device


vinc:
We can now configure the network manually from userspace:

net-dev

vinc:
The format for `/dev/net/usage` is: `<recv packets> <recv bytes> <sent packets> <sent bytes>`

```
~
> dhcp
[248.908097] NET IP 10.0.2.15/24
[248.916096] NET GW 10.0.2.2
[248.926094] NET DNS 10.0.2.3

~
> read /dev/net/usage
2 1180 2 620

~
> http example.com => /dev/null

~
> read /dev/net/usage
10 3312 10 1151
```

vinc:
The failing test on the previous commit due to memory allocation failure was weird.

vinc:
The device system could be refactored to be more hierarchical in the future instead of having just one big flat enum. For example all the `net-*` devices could be grouped into a virtual `net` device, sharing the same first byte on the device file, and then the second byte of the file would be used by the `net` module to dispatch the IO commands to the right IO NET device.

The ATA devices already use the next two bytes for the bus and drive numbers so there's a precedent for that.

vinc:
We'll need a way to monitor all network traffic and enable/disable the debug mode before removing the `net` command completely.

PR #702, opened 2024-11-14
Update HTTP server

vinc:
- [x] Add MIME type to BMP files
- [x] Use native newline in text content
- [x] Use native newline in HTML content


vinc:
HTTP/1.1 defines the sequence `CR LF` as the end-of-line marker for all protocol elements except the entity-body but until now the server would also convert the newlines in text files. A file created in one computer running MOROS and downloaded in another via HTTP would not have the same hash in both computer because the newlines would be different.

PR #704, opened 2024-11-27
Limit prompt line to screen width

vinc:
There's currently no multiline support in the prompt library (see #343 for a WIP) and a lot of weird things happen when a line goes past the screen width so the simplest fix for now is to prevent that.

prompt


vinc:
The fix handles line editing:

prompt2

vinc:
But it breaks down using file completion at the end of a long line:

prompt3

PR #706, opened 2024-11-28
Update clipboard behavior in editor

vinc:
The clipboard in the editor was a stack, leading to unexpected behavior. It will now be a single line buffer acting like the usual cut/copy/paste features. For example it will now be possible to copy a line and paste it multiple times.


vinc:
We could maybe move the cursor down after pasting a line to be on the new line, using exactly the same behavior as pushing the down key.

PR #708, opened 2024-12-04
Upgrade Rust to nightly-2024-12-01

vinc:
This is required to merge #707


vinc:
We need to upgrade the `x86_64` crate to compile without errors in this PR instead of #707.

PR #747, opened 2025-06-19
Add buffers to editor

vinc:
This PR will allow the edition of multiple files inside MOROS Editor.

- [x] Open new buffer (`Ctrl` + `O`)
- [x] Load next buffer (`Ctrl` + `Tab`)
- [x] Load previous buffer (`Ctrl` + `Shift` + `Tab`)
- [x] Kill current buffer (`Ctrl` + `K`)

The cursor position is saved when cycling between the opened buffers and the clipboard is shared globally to copy and paste between files.


vinc:
We can now open multiple files at once from the shell with `edit /tmp/*.txt` for example, or one by one inside the editor with `^O`.

PR #750, opened 2025-06-22
Add brainfuck interpreter

vinc:
I'm working on MOROS Shell 2.0 and I'll need a parser to handle the new language features. I don't like how hard it is with the `nom` crate to display good error messages with context for MOROS Lisp (which complicate a lot writing big programs inside MOROS), so I've been looking into the `chumsky` crate instead. That's the only issue I have with `nom` because it's a really good library otherwise.

When I learned how to use `nom` 4 years ago I added a small `calc` tool with a simple grammar (#263), so I'm now using the Brainfuck interpreter from the examples of the `chumsky` crate with the goal to extend it a bit and see if I like the crate.

I'm not sur if I'll merge this because I'm more interested in writing my own Brainfuck interpreter in MOROS Lisp, but I'm very interested in having a Brainfuck interpreter! Maybe I'll have both to compare the performance, like I did with the undocumented `pi` command (that I just remembered now).


vinc:
Hello world example:

```
~
> read hello.bf
++++++++[>++++[>++>+++>+++>+<<<<-]>+>+>->>+[<]<-]>>.>---.+++++++..+++.>>.<-.<.++
+.------.--------.>>+.>++.

~
> brainfuck hello.bf
Hello World!
```

Default error message:

```
~
> brainfuck error.bf
Error: found 'X' expected '<', '>', '+', '-', ',', '.', '[', or end of input
```

vinc:
Improved error message:

```
~
> brainfuck error.bf
Error: Unexpected token at error.bf:1:6

+++++X+++[>++++[>++>+++>+++>+<<<<-]>+>+>->>+[<]<-]>>.>---.+++++++..+++>
     ^ found 'X' expected '<', '>', '+', '-', ',', '.', '[', or end of input
```

vinc:
The hello world example has been taken from [Wikipedia](https://en.wikipedia.org/wiki/Brainfuck#Hello_World!) and the 400 bytes quine from [Daniel B Cristofani](https://brainfuck.org/400quine.b).

TODO: Add comment support for example attribution

vinc:
```
~
> read hello.bf
+++++
               +++[>++
     ++[>+
+>+
++>+
++>
+<<<<-]>+>+>->>+[<]<-]>>.>---.+++++++..+++.>>.<-.<.++
+.---X--.--------.>>+.>++.

~
> brainfuck hello.bf
Error: Unexpected token at hello.bf:8:6

+.---X--.--------.>>+.>++.
     ^ unexpected token
```

vinc:
Screenshot:

brainfuck

viktorpopp:
Why is the X an unexpected token? Shouldn't all text just be ignored?

vinc:
> Why is the X an unexpected token? Shouldn't all text just be ignored?

All tokens that are not instructions are now ignored but that was not the case with the example parser I started with from the crate's repo.

PR #752, opened 2025-06-26
Add better shell scripting capabilities

vinc:
At the moment we can only run one command at a time in the REPL and in any shell script, which is pretty limiting. This PR will evolve the shell into a proper scripting language inspired by Ruby allowing us to do much more with it:

- [ ] Add `;` to separate commands
- [ ] Add `and` operator
- [ ] Add `or` operator
- [ ] Add `not` operator
- [ ] Add `test` program
- [ ] Add `def <name> <args>* [do] <commands> end` (function definition)
- [ ] Add `if <condition> [do] <commands> [else <commands>] end` (conditional)
- [ ] Add `for <var> in <list> [do] <commands> end` (loop)
- [x] Add `$(<command>)` (command substitution)
- [ ] Add `<command> --> <command>` (pipe)

Example:

```ruby
def hello name
  print "Hello, $name!"
end

if test -e $USER
  hello World
else
  hello $USER
end

for file in $(list /tmp/*.tmp)
  drop $file
end
```

A real shell script could be a redefinition of `move` as `copy` + `drop` instead of being a dedicated Rust program:

```ruby
#! shell

def usage
  print "Usage: move <src> <dst>"
end

for arg in $*
  if test -e $arg "-h" or test -e $arg "--help"
    usage
    exit 0
  end
end

set src $1
set dst $2

if test -e $src "" or test -e $dst ""
  usage
  exit 1
end
        
copy $src $dst and drop $src
```

### DIFF
--- a/src/usr/shell/lexer.rs
+++ b/src/usr/shell/lexer.rs
@@ -1,0 +1,64 @@
+use alloc::vec::Vec;
+use core::fmt;
+use chumsky::prelude::*;
+
+pub type Span = SimpleSpan;
+pub type Spanned<T> = (T, Span);
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Token<'src> {
+    Null,
+    Str(&'src str),
+    Ident(&'src str),
+    Def,
+    Do,
+    End,
+    If,
+    Else,
+}
+
+impl fmt::Display for Token<'_> {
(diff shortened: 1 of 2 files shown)


vinc:
Command substitution has been implemented in #822

PR #754, opened 2025-06-29
Add higher level set macro to lisp

vinc:
MOROS Lisp had `var` to bind a value to a symbol and `set` to rebind another value to the symbol (causing an error if the variable didn't exist).

This PR will rename `set` to `mut` because it is a low level operation to mutate a variable, and create a new higher level `set` macro that either bind a value to a symbol with `var` if the variable doesn't exist or rebind the new value with `mut` otherwise.

This behavior add consistency between MOROS Lisp and MOROS Shell.

- [x] Rename `set` function to `mut`
- [x] Add `var?` function
- [x] Add `set` macro
- [x] Update doc
- [x] Update example scripts


vinc:
I'm working on adding `def` to MOROS Shell (#752) and with this PR they will both have the same `def` and `set` operations.

PR #761, opened 2025-07-14
Add spell command

vinc:
A spellchecker is pretty useful when you start to write a lot of text

PR #751, opened 2025-06-23
Improve panic handling

vinc:
I found that I could improve the panic handlers while working on error messages for the brainfuck interpreter in #750

PR #762, opened 2025-07-16
Add optional row and col params to path in editor

vinc:
We will now be able to open a file directly at the given coordinates with `edit <path>:<row>:<col>` following the convention set by the new error messages in #750, #751, and #761.


vinc:
The `row` and `col` params use 1-based indexing but the editor should not panic if one of them is zero.

PR #771, opened 2025-08-17
Replace bump frame allocator with bitmap frame allocator

vinc:
With this new frame allocator, we can now deallocate frames when they are no longer needed, fixing the memory leaks that occurred after process termination.

- [x] Replace bump frame allocator with bitmap frame allocator
- [x] Add frame deallocator
- [x] Deallocate process page table frame
- [x] Deallocate frame after unmap
- [x] Clean up page table after unmap
- [x] Compute usable frames once during init

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ qemu:
 
 test:
 	cargo test --release --lib --no-default-features --features serial -- \
-		-m $(memory) -display none -serial stdio \
+		-m $(memory) -cpu core2duo -display none -serial stdio \
 		-device isa-debug-exit,iobase=0xF4,iosize=0x04
 
 website:
--- a/Makefile
+++ b/Makefile
@@ -7,12 +7,12 @@ setup:
 	cargo install bootimage
 
 # Compilation options
-memory = 32
 output = video# video, serial
 keyboard = qwerty# qwerty, azerty, dvorak
 mode = release
 
 # Emulation options
+memory = 32
 smp = 2
 nic = rtl8139# rtl8139, pcnet, e1000
 audio = sdl# sdl, coreaudio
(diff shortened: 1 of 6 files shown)


vinc:
With the following lisp script:

```janet
(load "/lib/lisp/core.lsp")
(set i 1)
(set n 1024)
(while (<= i n) (do
  (sh (str "print hello " i " / " n))
  (set i (+ i 1))))
```

We could not run more than 192 times the `print` program with 32 MB of memory:

```
hello 188 / 1024
hello 189 / 1024
hello 190 / 1024
hello 191 / 1024
hello 192 / 1024
DEBUG: Could not allocate frame for Page[4KiB](0x4444c7d41000)
Error: Could not execute 'print'
Error: Panicked at src/sys/process.rs:336:13
       ^^^^^^^^ frame allocation failed
```

Now we can run it 738 times:

```
hello 733 / 1024
hello 734 / 1024
hello 735 / 1024
hello 736 / 1024
hello 737 / 1024
hello 738 / 1024
Error: Could not allocate 20298 bytes
```

Some memory leaks still occur but they are not related to deallocating frames after unmapping pages.

vinc:
There are indeed some optimization we can do to improve the performance but the time to boot and time to run a large program that does a lot of memory allocation is more or less the same with the new allocator.

vinc:
Using more RAM than usual with the new allocator increased the boot time, but now we are computing the usage regions only once during init and this reduce boot time a lot! For example with 2 GB of RAM the boot takes 0.76 s instead of 32.76 s.

vinc:
We used to need to cap the heap size to a low value even on real hardware because the frame allocation was so slow, not anymore!

PR #555, opened 2023-11-30
Add drive device

vinc:
- [x] Add `fs::DeviceType::Drive`
- [x] Implement `FileIO::read` for `ata::Drive` 
- [ ] Implement `FileIO::write` for `ata::Drive` 
- [x] Store bus number and drive number after the device type byte in the device file
- [x] Create device from its device file's whole buffer instead of only the first byte
- [x] Create `/dev/ata/<bus>/<dsk>` to read the drive device
- [ ] Add `SEEK` syscall
- [ ] Update `disk usage` command to read superblock from drive device
- [ ] Update `disk erase` command to write 0 to the whole drive device
- [ ] Update `disk format` to write 0 to the block bitmaps, create a new superblock, and create the root dir
- [ ] Update `disk list` to read more information from the drive device (name, size)

PR #772, opened 2025-08-25
Add FileIO::write to ATA

vinc:
Allow users to write sectors to hard drives directly via the device files in `/dev/ata`


vinc:
For a long time I wanted to try to copy the bootloader from memory at `0x7C00` and the kernel at `0x400000` and see if I could get a working boot image with it, but a few bits differs. We can get them elsewhere, for example we can find `31 C0` at `0xB420` to replace the first two bytes of the first sector of the loader (that are `00 00` in memory), this make the OS boot pretty far but not without issues.

```
> memory dump 0xB420 2 > /ini/kernel.img          # 0x31C0
> memory dump 0x7C02 53246 >> /ini/kernel.img     # Loader
> memory dump 0x400000 2244608 >> /ini/kernel.img # Kernel
> read /ini/kernel.img > /dev/ata/0/1
> disk format /dev/ata/0/1
```

Now that large downloads are possible it's going to be easier to just download an image of the kernel from the network instead of trying to rebuild it from memory.

vinc:
The kernel image is a bit too large to be downloaded directly so #773 will add zlib support to download a compressed version.

vinc:
Now we can do this to install the loader and the kernel image at the beginning of the disk:

```
> pkg install kernel
> read /ini/kernel.img > /dev/ata/0/0
```

vinc:
We could add a `zero` program or script that would continuously output `0` to stdin until interrupted to replace `disk erase /dev/ata/0/0` with `zero => /dev/ata/0/0`.

We could also do `print /dev/random => /dev/ata/0/0` but each buffer written to stdin should be a multiple of 512 bytes in length.

vinc:
This was a continuation of #555

PR #773, opened 2025-08-25
Add zlib support

vinc:
- [x] Add `inflate` command
- [x] Add `deflate` command
- [x] Add zlib support to `pkg` script


vinc:
On MOROS:
```
> deflate kernel.img
> inflate kernel.img.z
```

On Unix:
```
> pigz --zlib --best --suffix .z kernel.img
> unpigz kernel.img.z
```

We could also add `read -z` and `view -z` equivalent to `zcat` and `zless` on Unix.

src/usr/inflate.rs
```diff
@@ -0,0 +1,61 @@
+use crate::api::console::Style;
+use crate::api::fs;
+use crate::api::process::ExitCode;
+
+use alloc::format;
+use miniz_oxide::inflate::decompress_to_vec_zlib as inflate;
```
vinc, line 6: We'll want to switch to `decompress_to_vec_zlib_with_limit` at some point.

PR #776, opened 2025-08-27
Set VGA resolution

vinc:
The following resolutions can be written to `/dev/vga/mode`:
- `80x25c` (char mode)
- `320x200p` (pixel mode)
- `640x480p` (pixel mode)

TODO: Add `320x240p` and fix `640x480p`, both are planar modes while `320x200p` is a linear mode.

The format will evolve in time, for example it could become `320x200p 8bpp` and `640x480p 4bpp` without breaking changes. Or it could move to a ini file at the same time as all the other drivers.


vinc:
The code to reduce the kernel size and set the text mode during init was cherry picked from #835

vinc:
TODO: Add a write planar and a special clear if the resolution is 640x480x16

PR #777, opened 2025-08-27
Add geo.ini support to geo utilities

vinc:
The `geodate` and `geocal` utilities had multiple positional arguments: `<latitude>` for the second, `<longitude>` and `[<timestamp>]` for both. The second have two options, `--ephem` and `--solar`, while the first could use a `--format` option. They also didn't parse the `-h` or `--help` option.

I use them a lot, so I had to type my latitude and longitude a lot, and occasionally I would forget the order of the positional arguments.

This PR introduce the following changes:

- [x] Remove positional arguments
- [x] Add `-x` and `--longitude` option `geodate` and `geocal`
- [x] Add `-y` and `--latitude` option to `geocal`
- [x] Add `-t` and `--timestamp` option to `geodate` and `geocal`
- [x] Add `-f` and `--format` option to `geodate`
- [x] Add support for `/ini/geo.ini`

The format of `/ini/geo.ini` is the following:

```ini
lat = <float>
lon = <float>
```


vinc:
Usage:

usage

Example:

geo

PR #778, opened 2025-08-28
Add Intel I219 support in e1000 network driver

vinc:
- [x] Add Intel I219-V
- [ ] Add Intel I219-LM


vinc:
The `e1000` network driver works without changes on the Intel I219-V of my Lenovo ThinkPad X270 but not on the I219-LM of my T480.

PR #781, opened 2025-09-02
Add Sound Blaster 16 driver

vinc:
- https://wiki.osdev.org/Sound_Blaster_16
- https://pdos.csail.mit.edu/6.828/2006/readings/hardware/SoundBlaster.pdf
- https://github.com/qemu/qemu/blob/master/hw/audio/sb16.c


vinc:
We can now play a sound with either:

```
> read sound.pcm => /dev/snd/buf
> copy sound.pcm /dev/snd/buf
```

And stop it with either:

```
> read /dev/null => /dev/snd/buf
> copy /dev/null /dev/snd/buf
```

vinc:
Currently we can only play 22.1 kHz unsigned 8-bit mono PCM sounds.

They can be converted to the right format with the following command on Unix:

```
sox sound.mp3 -r 22050 -c 1 -b 8 -e unsigned-integer -t raw sound.pcm
```

The next step will be to add device file at `/dev/snd/ini` to set or get the current format.

vinc:
Using `sox sound.wav -r 44100 -c 1 -b 8 -e unsigned-integer -t raw sound.pcm gain -rh dither -s` produce a sound with much less noise while doubling the file size. It's a better compromise for this test.

vinc:
We can now use 8-bit mono WAV files in addition to raw 8-bit mono 44.1 kHz PCM files.

vinc:
Currently the convention is to use long names inside a driver directory so I renamed `/dev/snd/buf` to `/dev/snd/buffer` to match `/dev/vga/buffer`.

I'm also thinking about renaming `/dev/vga/mode` to `/dev/vga/config` and also merging `/dev/net/mac`, `/dev/net/ip`, and `/dev/net/gw` into `/dev/net/config` using the ini format to get or set a driver config. If we move to short names they could be renamed `/dev/<drv>/ini` to indicate the format, and then we also rename back buffers to `/dev/<drv>/buf`.

The config file for this sound driver could contain:

```ini
channels = 1
sample_bits = 8
sample_rate = 44100
```

And the matching video driver config could contain:

```ini
mode = 80x25
```

While the network driver config could contain:

```ini
mac = 52-54-00-12-34-56
ip = 10.0.2.15/24
gw = 10.0.2.2
```

vinc:
With this rewrite the device can keep track of the audio format it is playing which will help when 16-bit support is added next. The change to the module will also allow help supporting newer sound cards using AC'97 or Intel HD Audio later.

I stress tested the driver by streaming audio directly from the network without touching the hard drive. For that I downloaded a public domain recording of Chopin from here: https://archive.org/download/musopen-chopin/NocturneOp27No2.mp3

Then launched a TCP service in MOROS redirecting to the sound card:

```
> dhcp
> socket -l 80 => /dev/snd/buffer
```

And converted the MP3 in the host to a WAV file of the right format, then streamed it:

```
$ sh run/wav.sh /tmp/NocturneOp27No2.mp3 /tmp/chopin.wav
$ pv /tmp/chopin.wav | nc localhost 8080
```

PR #785, opened 2025-09-11
Add kernel package

vinc:
With this package we can copy the kernel and its loader to the beginning of the disk to boot MOROS from an ATA drive instead of a USB drive:

```
~
> dhcp
[3.038537] NET IP 10.0.2.15/24
[3.043537] NET GW 10.0.2.2
[3.052535] NET DNS 10.0.2.3

~
> pkg i kernel
Fetching '/var/pkg/kernel'
Fetching '/ini/kernel.img.z'
Inflating '/ini/kernel.img.z'

~
> read /ini/kernel => /dev/ata/0/0
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,10 @@ spell:
 pkg:
 	ls -1 dsk/var/pkg | grep -v index.html > dsk/var/pkg/index.html
 
+pkg-kernel:
+	cp $(bin) dsk/ini/kernel.img
+	sh run/deflate.sh dsk/ini/kernel.img
+
 clean:
 	cargo clean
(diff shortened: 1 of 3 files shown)


vinc:
The `Fetching` and `Inflating` actions have been renamed `Retrieving` and `Extracting` to have the paths aligned. The `.z` suffix has also be removed in the latter action to be more accurate. This will also work if we add `zip` support in addition to `zlib` because there will be potentially more than one file to extract from the archives:

```
> pkg i kernel
Retrieving '/var/pkg/kernel'
Retrieving '/ini/kernel.img.z'
Extracting '/ini/kernel.img'
```

PR #788, opened 2025-09-15
Add rfc package

vinc:
Add a `rfc` package containing a list of [Request for Comments](https://en.wikipedia.org/wiki/Request_for_Comments) documents describing the internet standards used by MOROS.


vinc:
This provide a good corpus of texts to test the text utilities in MOROS, in addition to offering self-hosted guidance on how to use some network protocols with the `socket` command. For example RFC 5321 describes the SMTP protocol to send mails for asynchronous messaging, and RFC 1459 describe IRC for synchronous messaging.

PR #792, opened 2025-09-18
Add mail command

vinc:
- [x] Add POP3 support
- [x] Add SMTP support
- [ ] Add POP3 over TLS proxy
- [ ] Add SMTP over TLS proxy


vinc:
The code repeats itself between the POP3 and the SMTP parts. I'm not sure yet if I want a `mail` command that dispatch the actions to the right protocol or lower level `smtp` and `pop3` commands.

PR #793, opened 2025-09-19
Store frame bitmap in memory region

vinc:
This PR stores the frame bitmap directly in memory in the first usable region big enough for it, instead of using a static array as implemented in #771 that forced us to limit the maximum amount of RAM to 4 GB in #775 to be able to boot MOROS with 16 or 32 MB of RAM in QEMU and still have enough memory left for userspace.

This is similar to what we do for the block bitmap in the filesystem.


vinc:
Tested successfully on a ThinkPad T440p with 16 GB of RAM, on a T480 with 40 GB, and on QEMU with as little as 12 MB of RAM.

PR #809, opened 2026-02-14
It may be little thing but it's just a small typo if it is not that good to merge just del the pr

udaykiriti:
Its just a small type.
I dont know is it intentional or mistake..

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,21 +20,21 @@ be accomplished after completing an OSDev tutorial, while reading the
 ## Features
 
 - External bootloader (using [bootloader][4])
-- x86 CPU support (using [x86_64][5])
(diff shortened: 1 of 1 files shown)


viktorpopp:
Intentional

udaykiriti:
Its my bad , in my code editor it shows like red warning for this thing  "_" thing so i thought that it is some mistake . So by mistake i removed that underscore..


It's totally my fault...

vinc:
No worries :)

I see you're also trying to contribute to bigger projects like Hermit and Serenity, that's pretty cool! I'd suggest checking their issues page and picking known bugs to fix. The list on Serenity is long given the size of the OS, and you can have really impactful contributions that way!

PR #817, opened 2026-03-27
Improve editor search

vinc:
- [x] Add backward search with Ctrl+Shift+N to complement forward search with Ctrl+N
- [x] Replace string search with regex search
- [x] Add word movement with Ctrl+Left and Ctrl+Right


vinc:
Okay we now have proper word jumping in the editor!

PR #820, opened 2026-03-30
Reuse deleted dir entries

vinc:


vinc:
Previously a deleted dir entry was just marked as deleted by clearing its address bytes, but dropping a lot of files from a dir could end up wasting a lot of unused space in the directory blocks.

Now when an entry is deleted we shift the following entries of the block over it and clear the space created at the end of the block. Later when a new entry is created we can reuse this space if it's big enough to hold the entry. In order to do this we create a virtual dir entry returned by the read dir iterator to mark this space as unused.

We also try to reuse a deleted dir entry from previous versions of MOROS but we can do this only if the new entry is exactly the same size as the previous one otherwise we would corrupt the read dir iterator by leaving a gap before the next entry.

TODO: Remove a block from the linked list after deleting its last dir entry if it's not the first one.

vinc:
I also looked into shifting all the remaining entries of the dir, or finding one or more subsequent entries that could fit at the end of the current block, possibly starting from the end to avoid shifting things too much. This could be done either during deletion, or periodically as a form of compaction to keep using space as efficiently as possible.

In the end I'm only doing it in the current block because this operation is mostly free, we needed to write the block anyway when clearing the address of the deleted entry so it won't be slower to shift and clear more bytes in the memory buffer before the slow ATA write.

PR #790, opened 2025-09-16
Add draw command

vinc:
Add command to draw text on screen in VGA graphics mode.

This PR is another tiny step toward adding UEFI support with the goal of one day maybe supporting `ARM64` and `RISCV64` in addition to `AMD64`.

PR #822, opened 2026-04-01
Add pipe device file

vinc:
This PR implements a pipe with a shared buffer acting as a queue accessible at `/dev/pipe`. When a file handle to a pipe is duplicated, both handles share the same buffer. Data written to one handle can be read from the other, and reading consumes the data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Add pipe device file (#822)
 - Fix potential deadlocks (#824)
 - Reuse deleted dir entries (#820)
 - Improve editor search (#817)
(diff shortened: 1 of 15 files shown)


vinc:
In MOROS Lisp the `(shell <cmd>)` function returns the exit status of the command, and now we can capture the output of the command with `(shell->binary <cmd>)`:

```
~
> lisp
MOROS Lisp v0.7.0

> (str/trim (bin->str (shell->binary "date")))
"2026-04-01 20:09:20 +0200"
```

vinc:
When the `draw` command was introduced in #790 with the option to display the result of a command at periodic intervals we used a redirection to a temporary file, which resulted in changes needed to better handle file deletions in #820, but now we can skip the temporary file and just use a pipe.

vinc:
And now, command substitutions in the shell!

```
~
> copy hello.txt hello-$(date %Y%m%d)-$(date %H%M%S).txt

~
> list
14 2026-04-01 18:01:42 hello-20260401-180142.txt
14 2026-04-01 17:12:52 hello.txt
```

PR #823, opened 2026-04-02
Refactor syscall handler and add FileList, CreateDir, Version syscalls

BogdanUser404:
I have performed a small refactoring of the syscall dispatcher by replacing the constants in number.rs with a type-safe 

- SyscallCode enum. I also renamed the variable n to syscall_code for better clarity.
- Added 3 new system calls:
- Version: Returns a packed version ID.
- CreateDir: Creates a directory at the specified path.
- FileList: Populates a provided buffer with the list of file names and returns the number of entries.

This is one of my first Pull Requests to this project. I've also updated the Makefile to allow testing on different CPU architectures and memory configurations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,81 +1,3 @@
-# USAGE GUIDE:
-# Use 'make image' to build the OS disk.
-# Use 'make qemu' to run the emulator with default settings.
-# Use 'make qemu cpu=athlon64 memory=128 smp=1' to override settings.
-# Use 'make qemu kvm=true' for hardware acceleration (Linux only).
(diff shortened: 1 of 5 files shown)


vinc:
Hey, thanks for the PR! I won't be merging it though. Creating a directory and listing its contents are already covered by the `OPEN` and `READ` syscalls, so there's no need for dedicated ones. I try to keep the syscall count as low as possible and favor generic file operations.

More generally, I prefer PRs that fix bugs rather than add features. That said, I built MOROS to be forked and experimented with, so please do play with it!

PR #826, opened 2026-04-04
Improve disk caching

vinc:


vinc:
This change update the cache on writes instead of clearing it. The `pkg install rfc` command went from 53 to 45 seconds.

vinc:
Use a vector stored on the heap instead of an array stored on the stack. With this change we'll also be able to change the cache size dynamically at runtime.

PR #828, opened 2026-04-06
Add box to VGA palette array

vinc:
This PR fixes the following issue:

```
warning: large size difference between variants
   --> src/sys/fs/mod.rs:130:1
    |
130 | / pub enum Resource {
131 | |     Dir(Dir),
    | |     -------- the second-largest variant contains at least 48 bytes
132 | |     File(File),
133 | |     Device(Device),
    | |     -------------- the largest variant contains at least 776 bytes
134 | | }
    | |_^ the entire enum is at least 776 bytes
    |
help: consider boxing the large fields or introducing indirection in some other way to reduce the total size of the enum
   --> src/sys/fs/mod.rs:133:5
    |
133 |     Device(Device),
    |     ^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
```

### DIFF
--- a/src/sys/fs/block_device.rs
+++ b/src/sys/fs/block_device.rs
@@ -86,7 +86,7 @@ impl BlockDeviceIO for MemBlockDevice {
 }
 
 pub fn mount_mem() {
-    let mem = sys::mem::memory_free() / 3;
+    let mem = sys::mem::memory_free() / 2;
     let len = mem / super::BLOCK_SIZE; // TODO: take a size argument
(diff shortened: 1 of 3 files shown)


vinc:
The only limitation is that we can't set the palette during driver initialization as it's done before alloc is available.

vinc:
We can now set the size of the in memory filesystem back to 1/2 of the RAM instead of 1/3 when running the tests and keep adding more device files. See #822 and probably #781 too.

PR #829, opened 2026-04-11
Add AC'97 driver

vinc:
Links:
- [OSDev Wiki](https://wiki.osdev.org/AC97)
- [Intel Spec](https://github.com/user-attachments/files/26649208/ac97_r23.pdf)


vinc:
I can now play 16-bit stereo sounds but interrupts are not firing.

vinc:
Interrupts are now firing!

We can stream 16-bit stereo from the network but no more than 11025 Hz for now. There's also a big memory leak somewhere.

vinc:
I tried to make the driver work on real hardware but the only PC I have with AC'97 is based on an ATI SB400 and a Realtek ALC655 that looks very different from the Intel ICH in QEMU.

vinc:
This spec is better for implementing the driver: https://theretroweb.com/chip/documentation/30701701-6803f2f676107531879811.pdf

PR #831, opened 2026-04-14
Split process module

vinc:
This PR split the process module into multiple files.


vinc:
This module has always been pretty experimental and not well organized. The most obvious split is to put everything regarding `Process` in one file and `ProcessTable` in another file. After that we can extract `Process::spawn` into a `Spawn` trait in its own file.

vinc:
The `Spawn` trait now only exposes `Process::spawn`, which internally calls `create` and `exec` without leaking them to the rest of the codebase.

vinc:
The code extracted from `exit()` into `load_process()` and from `exec()` into `copy_args()` will make it easier to do a cooperative multitasking experiment later on.

PR #836, opened 2026-04-21
Add process devices

vinc:
Add process device files to change process data from userspace:

- [x] `/dev/proc/id` to get the PID
- [x] `/dev/proc/dir` to get or set the current dir
- [x] `/dev/proc/env` to get or set env vars
- [x] `/dev/proc/user` to get the current user


vinc:
We can't use `print` to change the current dir in the shell as it's an external program, but an `echo` builtin will do:

```
~
> echo /tmp => /dev/proc/dir

/tmp
> echo $dir
/tmp
```

The variable to get the current dir in the shell changed from `$DIR` to `$dir` as it is a shell variable and not an environmental variable.

vinc:
I removed `env` and replaced it by a new option `--env` for the shell builtin `set`:

```
> read /dev/proc/env
HOME = "/usr/vinc"
USER = "vinc"

> echo TEST=42 => /dev/proc/env

> read /dev/proc/env
HOME = "/usr/vinc"
TEST = "42"
USER = "vinc"

> date
2026-04-21 19:44:24 +0000

> set --env TZ 7200

> read /dev/proc/env
HOME = "/usr/vinc"
TEST = "42"
TZ   = "7200"
USER = "vinc"

> date
2026-04-21 21:44:43 +0200
```

vinc:
TODO:
- How to unset an env var?
- How to login from userspace?
- How to keep `$dir` up to date in the shell?
- Replace `print` program with a shell builtin?

PR #841, opened 2026-04-27
Improve Lisp user experience

vinc:
- [x] Use only short function names
- [x] Add location to parse and eval errors
- [x] Make `and` and `or` variadic
- [x] Merge `=` for numbers into generic `eq?` (aliased to `=`)
- [x] Add `num/eq?` for stricter number comparisons
- [x] Rename `trunc` to `num/int`
- [x] Change `socket/accept` to return `()` instead of an error
- [x] Remove `(slice index enumerable)` to keep only `(slice index length enumerable)`
- [ ] Add script to copy docstrings to doc
- [ ] Add script to execute examples and compare the return values after `# =>`
- [x] Move `head` and `tail` from special forms to alias of `first` and `rest` in core lib


vinc:
The great shortening has happened! For the longest time I couldn't decide between long (`define`) and short (`def`) function names. In practice I use short names everywhere, even if short names where aliases of the default long names. So short function names win.

vinc:
This commit add locations to parse and eval errors, this is specially useful when the error occur in another file!

Parse error in another file:
```
> lisp parse-error.lsp
Error: Could not parse ') # Error' at /lib/lisp/math.lsp:11
Error: Could not load file at /lib/lisp/core.lsp:164
Error: Could not load file at parse-error.lsp:1
```

Eval error in another file:
```
> lisp eval-error.lsp
Error: Could not find symbol 'nope' at /lib/lisp/math.lsp:11
Error: Could not load file at /lib/lisp/core.lsp:164
Error: Could not load file at eval-error.lsp:1
```

vinc:
I started documenting all the functions with examples over the weekend, it's slow going but quite needed as I often forget some little details here and there already.

TODO: turn the examples into a test suite

vinc:
Documenting `socket/accept` reminded me that this syscall could be improved. It's only used by the `socket` command currently because the `httpd` command still talk directly to the network card, bypassing the syscalls.

It is blocking and returns either an error after a timeout or the remote address. Instead it should return an error or a tuple with a new file handle along with the remote address and port. That way we can close this new connection handle when done and keep the listening handle open. Instead of blocking we should poll the listening handle that should tell when there's a new connection to accept.

The function will then return either `()` or `(handle address port)`. I already replaced the error with `()` because truthy value in case of success is a better pattern here.

vinc:
TODO: Move `head` and `tail` to core lib instead of special forms

vinc:
Looking at `cond`:

```
(cond
  ((> x 0) "positive")
  ((< x 0) "negative")
  ((= x 0) "zero")
  (true "nan"))
```

Then `case`:

```
(case x
  (1 "one")
  (2 "two"))
```

Then `let`:

```
(let (x y) (1 2) (+ x y)) # => 3
```

And finally `dict`:

```
(dict/pairs (dict "a" 1 "b" 2)) # => (("a" 1) ("b" 2))
```

I think `dict` could become `(dict ("a" 1) ("b" 2))`.

And `let` could become `(let ((x 1) (y 2)) (+ x y))` or `(let (x 1) (y 2) (+ x y))`.

If I don't change `dict` then `let` could also be `(let (x 1 y 2) (+ x y))`.

This require some thinking and will not be a part of this PR though.

PR #848, opened 2026-05-12
Improve Lisp binding

vinc:
- [x] Move `eval::eval_args` calls out of `env::inner_env` in interpreter
- [x] Remove `env::function_env` and `env::macro_env` in interpreter
- [x] Rename `env::inner_env` to `env::bind` in interpreter
- [x] Allow optional arguments
- [x] Add `eval::apply` in interpreter
- [x] Add back `apply` special form without double eval bug
- [x] Add `fold` special form
- [x] Update doc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Improve Lisp binding (#848)
 - Add process devices (#836)
 - Improve Lisp user experience (#841)
 - Add assembler written in Lisp (#839)
--- a/doc/lisp-doc.md
+++ b/doc/lisp-doc.md
@@ -108,12 +108,23 @@
 ### quote '
 
     '(1 2 3) # => (1 2 3)
+
 ### quasiquote `
 ### unquote ,
 ### unquote-splice ,@
+
 ### splice @
 
+    ((fun (a b @c) c) '(1 2 (3))) # => (3)
+    ((fun (a b @c) c) '(1 2 ()))  # => ()
+    ((fun (a b @c) c) '(1 2))     # => ()
+
 ## Meta
+
+### apply
+
+    (apply + '(1 2 3)) # => 6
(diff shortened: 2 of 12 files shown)


vinc:
If we dissociate function arguments evaluation from parameters binding we don't need a special case for macros, so we can merge `function_env` and `macro_env` into `bind` and simplify the code.

vinc:
We couldn't have optional args before:
```
> ((fun (a b @opts) opts) 1 2 3 4)
(3 4)

> ((fun (a b @opts) opts) 1 2)
Error: Expected at least 3 arguments, got 2
```

This is now fixed:
```
> ((fun (a b @opts) opts) 1 2 3 4)
(3 4)

> ((fun (a b @opts) opts) 1 2)
()
```

vinc:
This last commit change the evaluation order but I think the new order is fine:

Before:
```
> (+ 1 2 3)
6

> (a 1 2 3)
Error: Could not find symbol 'a'

> (1 2 3)
Error: Expected first argument to be a function

> (1 2 a)
Error: Expected first argument to be a function
```

After:
```
> (+ 1 2 3)
6

> (a 1 2 3)
Error: Could not find symbol 'a'

> (1 2 3)
Error: Expected first argument to be a function

> (1 2 a)
Error: Could not find symbol 'a'
```

vinc:
Now that `bind` is no longer evaluating the function arguments it is possible to reintroduce the `apply` special form without the double eval issue from before.

This works now:
```
> (apply (fun (x) x) (list '(1 2)))
(1 2)
```

vinc:
With the `apply` helper it was straightforward to add a (left) `fold` special form:
```
> (fold - 0 '(1 2 3))
-6
```

vinc:
With `fold` we can now fix `reduce`:
```
> (reduce - '(1 2 3))
-4

> (reduce - '(1 2))
-1

> (reduce - '(1))
1

> (reduce - '())
()
```

PR #849, opened 2026-05-13
Remove chumsky crate

vinc:
- [x] Rewrite INI parser with Nom
- [x] Rewrite Brainfuck parser with Nom


vinc:
Last year I tried to replace `nom` with `chumsky` for all my parsing needs.

I started by implementing a small Brainfuck interpreter in #750 then a INI config parser in #759 but for the `spell` command it turned out to be easier to write a hand parser than using `chumsky`.

I realized that I preferred `nom` better to keep the complexity low by combining many small functions. I should probably look into `nom_locate` to get back what I'll miss from `chumksy` but it turns out that I can do more than what I though to locate errors out of the box.

PR #843, opened 2026-05-01
Improve process memory

vinc:
This PR create the process page table from scratch, sharing only what is needed instead of cloning every tables. The kernel heap area and the processes area (in the kernel page table) were moved to different L4 indexes to simplify this.

The processes area in the kernel page table located high in memory is also remapped low in the process page table where programs are linked, to avoid a hack in the page fault handler where this was done as demand paging backed by memory, that is by copying the program code from high memory during instruction fetching.

PR #869, opened 2026-07-14
Improve process memory isolation

vinc:
This PR is a continuation of #843 but instead of trying to build a process page table from scratch we keep the idea of cloning the kernel page table but we move user memory to a dedicated L4 entry that is not mapped inside the kernel. This way each process is isolated and everything is simplified.


vinc:
This PR was the last step before running DOOM inside MOROS and an important cleanup on the road to MOROS 1.0 so I'm pretty happy about it!

Next step after DOOM will be to move the args to the process stack instead of being in the middle of its memory.

PR #879, opened 2026-08-08
Remove dependency on x86_64::instructions::random

vinc:


vinc:
The `Makefile` has to be changed to `-cpu IvyBridge` instead of `-cpu core2duo` to test this PR with `RDRAND` support.

Given that the feature became available in 2012 for Intel and 2015 for AMD the instruction should probably move to a `x86_64` or `amd64` module instead of the generic `x86` that will also work for the i686 port. Although it will be straightforward to add a compilation option to return `None` on i686.

PR #835, opened 2026-04-18
Add Limine support

vinc:
This PR is quite experimental! The goal is to see what is needed to support Limine with its native protocol and/or multiboot2.

Limine with its native protocol doesn't support VGA Text Mode but it places the kernel in higher half memory, which is interesting and something that the bootloader crate can't do with the 0.9 version.

Limine with the multiboot2 protocol does support VGA Text Mode, but doesn't enable long mode.

PR #889, opened 2026-08-16
Add Grub support

vinc:
This is a continuation of the work done in #835 that added partial Limine support with `limine` and `multiboot2` protocol and the beginning of the i686 port.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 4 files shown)


vinc:
TODO:
- Reduce image size
- Remove timeout

vinc:
I couldn't get Limine to boot with multiboot2 on my old test computer but Grub does. It is quite heavier but it's an interesting alternative to support.

PR #897, opened 2026-08-21
Replace x86_64 GDT and TSS structures

vinc:
Another PR removing more usages of the `x86_64` crate for the i686 port.

I'm using the code of both `x86_64` and `x86` crates along with the OSDev wiki as a starting point. The crates must be generic to work for a lot of projects but I just need a subset of the features for MOROS so there's room for simplification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Replace x86_64 GDT and TSS structures (#897)
 - Add Grub support (#889)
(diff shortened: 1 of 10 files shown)


vinc:
The Intel and AMD manuals were the best sources in the end to find all the bits to set in the segment descriptors:

- https://cdrdv2.intel.com/v1/dl/getContent/671447
- https://docs.amd.com/v/u/en-US/24593_3.45_APM_Vol2

PR #899, opened 2026-08-24
Add VGA text mode support for i686

vinc:


vinc:
The PR #888 changed the VGA init to set the text mode explicitly and removed the call to clear the screen, probably because there's also one inside the text mode function, but the latter clear the screen in graphics mode while the former does it in text mode, so I restored that.

Initially this PR had to reduce the scroll buffer size to avoid overflowing the stack but this can be fixed by using static buffer so I implemented that. Although I'll open a PR to set the same stack size explicitly for the three supported bootloaders because right now they all have a different size.

There was also an off by one error in the code reducing by one line the usable length of the scroll back buffer.

vinc:
Kernel size with the commit adding the non-zeroed static buffer: 2328064 bytes
Kernel size before that, and after the latest commit using a zeroed buffer: 2288640 bytes
Difference: 40 KB, the size of the buffer that moved from `.data` to `.bss` in the image.

PR #901, opened 2026-08-24
Set the kernel stack size

vinc:
- [x] Rust bootloader crate
- [x] Limine protocol
- [x] Multiboot2 protocol
- [x] TSS module


vinc:
This could have fixed the scroll buffer init issue in #899 on multiboot2 because the kernel stack was too small, there were a better fix for that but setting the same size explicitly for each bootloader is a good thing to do.

PR #905, opened 2026-08-28
Add console::Input for i686

vinc:
A `String` was used for `STDIN` which required an allocator quite early in the boot process, this PR replaces it with a fixed buffer with a special slot at the end for the newline char.


vinc:
With the console compiling on i686 it's now possible to enable serial input.

PR #907, opened 2026-08-30
Enable tests on i686

vinc:


vinc:
The code is a bit of a mess with the three bootloader protocols and two architectures but there's now a Limine test runner for i686 with multiboot2!

vinc:
Moving all the functions related to the tests in a `test.rs` module instead of `lib.rs`, some of which that are similar than the ones in `main.rs` reduce the above mentioned mess significantly.

PR #909, opened 2026-08-31
Enable cpuid on i686

vinc:


vinc:
The `raw_cpuid` crate requires the `sse` feature on `x86`, which we don't have inside the kernel, to use `CpuId::new()`, even on a CPU that has SSE like a Pentium III, so I had to duplicate the implementation without the `cfg` guard (see https://github.com/gz/rust-cpuid/issues/134).

It works:

```
Loading MOROS ...
[0.260960] CPU GenuineIntel
[0.267959] CPU Intel(R) Atom(TM) CPU N270   @ 1.60GHz
[0.275958] MEM [0x00000000000000-0x0000000009FBFF] Usable
[0.277958] MEM [0x0000000009FC00-0x0000000009FFFF] Reserved
[0.278958] MEM [0x000000000F0000-0x000000000FFFFF] Reserved
[0.278958] MEM [0x00000000100000-0x00000000166FFF] Kernel
[0.279957] MEM [0x00000000167000-0x00000001FDFFFF] Usable
[0.279957] MEM [0x00000001FE0000-0x00000001FFFFFF] Reserved
[0.279957] MEM [0x000000FFFC0000-0x000000FFFFFFFF] Reserved
Loaded MOROS successfully!
```

I also had to use `n270` instead of `pentium3` as CPU on QEMU because the latter doesn't have a processor brand string.

PR #912, opened 2026-09-01
Add sys::x86::addr module

vinc:
- [x] Add `sys::x86::addr::PhysAddr`
- [x] Add `sys::x86::addr::VirtAddr`
- [x] Enable `sys::acpi`
- [x] Enable `sys::snd`
- [x] Enable `sys::net`


vinc:
I extracted the commits about `api::hex` to #913

PR #916, opened 2026-09-05
Use usize milliseconds instead of float seconds for sleep

vinc:
We can't pass a `f64` directly to a syscall on i686 and I've never been satified with using a float for the sleeping duration at the low level so this PR changes that. The sleep program and lisp still use a number of seconds in float at the high level.

- https://9p.io/magic/man2html/2/sleep
- https://en.wikipedia.org/wiki/Sleep_(system_call)


vinc:
TODO: benchmark `httpd` with `siege` to measure the impact of clamping the sleep duration, and see if we can reduce CPU usage when idle.

vinc:
DOOM needed to be updated too (https://github.com/vinc/moros-doom)

PR #921, opened 2026-09-11
Enable paging on i686

vinc:
TODO:
- [x] Add `cpu::has_pse()`
- [x] Add `x86::regs::Cr0`
- [x] Add `x86::regs::Cr4`
- [x] Add `x86::page::PageTable`
- [x] Add `x86::page::PageTableEntry`
- [x] Add `x86::page::PageTableFlags`
- [x] Add `mem::paging::init()`
- [x] Add `mem::mapping`

LINKS:
- https://en.wikipedia.org/wiki/Control_register#Control_registers_in_Intel_x86_series
- https://en.wikipedia.org/wiki/Page_Size_Extension

### DIFF
--- a/src/sys/cpu.rs
+++ b/src/sys/cpu.rs
@@ -39,8 +39,20 @@ pub fn init() {
     }
 }
 
-// RDRAND has been available since 2012 for Intel (Ivy Bridge) processors
-// and 2015 for AMD.
+// RDRAND (Read Random)
+//
+// Support:
+// - Intel Ivy Bridge (2012)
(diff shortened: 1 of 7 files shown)


vinc:
I created a `sys::x86::page` module for the new page table structures that are used in `sys::mem` on i686 and there's `sys::mem::paging` for AMD64 that could be mostly moved to `sys::mem::mapping` to make room for the new paging code.

I'm still figuring out what I need, but like the rest of the i686 port, the structures are both simpler and less generic than what I was using from the `x86_64` crate on AMD64.

PR #922, opened 2026-09-11
Add i586 support

vinc:


vinc:
Some slightly unrelated changes on that branch in the past few commits, but I'm testing the OS on old hardware and fixed a few issues, most notably the ATA identify command that wasn't waiting correctly on an IBM ThinkPad X20 (i686).

